PR #146, opened 2026-08-31
An entity with no fact attached tells the reader nothing

WaylandYang:
The last few rounds were all about whether what gets extracted is **correct** (reversals, violations, fake entities). But a 0% violation rate can be bought by extracting less — so there needs to be an opposing measure: **isolates**, entities with no fact attached.

## Two guesses ruled out first

**Not chunking.** 14.7% of isolates have their mention in the last 20% of a chunk, against 15.0% for entities with edges — nearly identical distributions. If relations were being cut across chunk boundaries, isolates should skew towards the edges.

**Not invented names.** Of 73 isolates, **zero** have names that cannot be found in the text; all of them are there verbatim.

## The real answer: seven in ten are missed extractions

Sampling the source text:

```
Vinod Khosla    "venture capitalist Vinod Khosla expressed interest in Altman's potential venture"
Grok-4          "On July 9, 2025, xAI unveiled Grok-4"
Satya Nadella   "Microsoft chief executive Satya Nadella" + "investing US$10 billion into OpenAI"
UNESCO          "Hugging Face teamed up with Meta and UNESCO"
Voxtral TTS     "In March 2026, Mistral AI released Voxtral TTS"
```

The model recognised the entity and did not extract the relation **stated plainly in the same sentence**. `xAI unveiled Grok-4` and `Mistral AI released Voxtral TTS` are exactly the product launches that should be extracted.

`Satya Nadella` makes the point best: he appears in two chunks with an explicit relation in each, and has no edges at all — which "only mentioned once" cannot explain.

## The fix: one prompt rule

> Everything you list in entities should appear in at least one fact. If the text does not let you state a relation about it, do not list it in entities — a name with no fact attached tells the reader nothing. Check each one before finishing: an entity in no fact means either you missed a relation the text states, or it should not have been listed.

## Numbers (same corpus and ontology, 60/60 chunks)

| | facts | entities | **isolates** | no predicate | violations |
|---|---|---|---|---|---|
| without the rule | 427 | 417 | 73 = **17.5%** | 199 | 0 |
| **with the rule** | 454 | 383 | 49 = **12.8%** | 250 | 0 |

**It tightens from both ends**: 34 fewer entities (no longer listing things it has nothing to say about) and 27 more facts (picking up the relations it was missing). Violations remain 0.

The cost is predicate-less facts going 199 → 250 — some of the 27 extra facts have no ontology relation to land on, so they become a predicate-less fact plus the source wording. That is honest silence, not a loss.

## It does not solve the underlying problem

12.8% means a substantial number are still being missed. One line of prompt catches some of them, not all. **Isolates are now a yardstick**, and the next round should go after why the rest are still missed.

Worth recording in passing: a mid-run snapshot at 17 chunks measured 6.0% against 12.8% for the full run — **a partial snapshot is not a result.**

Design record 0012 gains a seventh section.


WaylandYang:
撤回。规则 7a 确实把孤点从 17.5% 降到 12.8%，但抽样看恢复出来的事实发现代价不可接受——**它逼模型为每个实体凑事实，凑不出来就编**：

```
The New York Times --diversity_staffing_report-> OpenAI
    原文：「The removal left OpenAI in 'chaos', according to The New York Times.」
Jason Kwon --optimistic_about_return-> Sam Altman
    原文：「OpenAI was optimistic it could return Altman, Brockman, and other employees」← 主语是 OpenAI
```

`diversity_staffing_report` 跟原文毫无关系，是杜撰；第二条主语张冠李戴。

**孤点只是没说话，杜撰是说了假话。** 这跟这一整轮的方向（宁可沉默不要自信的错误）正相反，所以不合。

顺带查出一个独立的缺口，另行处理：**空谓词事实的方向完全没人管**——签名掰正只作用于有本体谓词的事实，而空谓词占 55%。恢复出来的 `Grok-4 --software_add_on-> xAI`（原文是 xAI 发布 Grok-4）就是这一类。

PR #201, opened 2026-09-02
An empty balance is final, so the queue should stop asking

### LINKED ISSUE #195: An empty balance should not spend the retry budget
A job that fails on `OutOfCredit` still spends its three attempts at 30s, 120s and 270s. A balance does not refill in seven minutes, so the retries only restate the same error and delay the failure the operator needs to see.

`RateLimited` deserves those retries; #176 separated the two types precisely because one recovers on its own and the other does not. The retry policy has not caught up with that split.

Either skip retries for `OutOfCredit`, or give the job table a way for a handler to say "this one is final".

WaylandYang:
Closes #195.

## What

A job that fails on `OutOfCredit` spent all three attempts — 30s, 120s, 270s. No balance refills in seven minutes, so the retries restated the same error three times and delayed the failure the operator needed to see by seven minutes. `RateLimited` deserves those retries, and #176 separated the two types precisely because one recovers on its own and the other does not; the retry policy had not caught up.

## How

The issue offered two routes. Skipping retries for `OutOfCredit` inside the queue is not available: `utopia-store` does not depend on `utopia-llm` and should not — the queue has no business knowing what a model endpoint is. So this takes the second: a handler can say "this one is final".

`utopia_core::Terminal` is a marker error. The worker's handler wraps a hopeless failure with it, `retry_delay` sees it and returns `None`, and the job goes to `failed` on the first attempt. Any handler in any crate can use it; the queue only asks whether the marker is there.

`alerting::hopeless` decides what qualifies, and only `OutOfCredit` does. It reads the same typed error `alert_for` reads, so "reported as an error" and "not retried" stay the same class and cannot drift apart. A rate limit still gets its backoff; so does an ordinary parse failure, because one network hiccup should not be fatal.

One thing that had to move with it: `observe_job_failure` reported only when `attempts >= max_attempts`. Failing on the first attempt would have slipped past it — faster failure and nobody told, which is worse than before. It now also reports when the failure is marked terminal.

Also fixed in passing: the retry path bound the `Job` struct where it meant the error string, and the failure path wrote `e.to_string()` rather than the full context chain.

## Checks

`cargo clippy --workspace --all-targets` clean. `retry_delay` is a pure function with two cases (ordinary backoff and give-up, terminal stops at once); `hopeless` has three (empty balance is final, rate limit is not, ordinary failure is not). Full workspace suite green against a scratch database.

### DIFF
--- a/crates/utopia-core/src/error.rs
+++ b/crates/utopia-core/src/error.rs
@@ -50,3 +50,33 @@ impl AppError {
 }
 
 pub type AppResult<T> = Result<T, AppError>;
+
+/// 标在一个**不会因为重试而变好**的失败上（见 issue #195）。
+///
+/// 队列的默认假设是「再等一会儿也许就好了」，多数失败确实如此：端点抖一下、
+/// 数据库忙一瞬、限流一分钟就过去。余额耗尽不是——三次重试隔着 30 秒、2 分钟、
+/// 4 分半，七分钟里没有人会去充值，重试只是把同一句错误重说三遍，而运维需要
+/// 看见的那条「失败」被推迟了七分钟才出现。
+///
+/// **判据留在处理器那一侧，不在队列里。** 什么算没救跟领域有关——
+/// `utopia-store` 看不见 `utopia-llm` 的错误类型，也不该看见。处理器把这个标记
+/// 挂上去（`err.context(Terminal)`），队列只问「挂了没有」。
+///
+/// 挂上它不影响别的：告警照报（`observe_job_failure` 一并认这个标记，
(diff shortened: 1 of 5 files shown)


WaylandYang:
One cost worth stating before this merges: a batch. If credit runs out halfway through 500 queued documents, the old behaviour retried each one over ~7 minutes, so a top-up inside that window rescued the rest by itself. Now every remaining job is `failed` at once and has to be triggered again after the top-up — per document that is a click, across a batch it is the whole batch.

Taken deliberately: the operator learns the real state seven minutes sooner, and re-running failed jobs is a gap that predates this change. Filed as #216.

PR #204, opened 2026-09-02
docs(readme): describe reasoning accurately

Floating-Y:
## Summary

Describe Utopia's reasoning pipeline without implying that it exposes a user-facing rule language.

## Context

The current README states that rules are expressed in temporal Datalog. As noted in #163 and documented in [ADR 0002](https://github.com/deeplethe/utopia/blob/dev/docs/decisions/0002-reasoning-engine.md), ontology axioms compile into internal rules; users do not author rules directly.

## Changes

- State that ontology axioms compile into rules evaluated by forward chaining.
- Preserve the existing explanation of derived-fact validity, provenance, and derivation paths.
- Remove the inaccurate implication of a user-facing temporal Datalog language.

## Scope

This PR changes one row of the feature table in `README.md`. It is intentionally independent of the other README follow-ups.

## Validation

- `git diff --check`
- Documentation-only change; no runtime tests required


WaylandYang:
Thanks for this. The Features table was rewritten in #217 and that row now reads "Ontology axioms compile into rules: transitivity, symmetry, inverses and relation hierarchy derive new facts by forward chaining" — no temporal Datalog claim any more, which is what this PR set out to fix. Closing as superseded.

PR #205, opened 2026-09-02
docs(readme): clarify browser-based interfaces

Floating-Y:
## Summary

Clarify that the system console, graph browser, and ontology workbench are all browser-based interfaces.

## Context

In the current sentence, the relative clause “that runs in the browser” appears to modify only the ontology workbench. This change addresses the ambiguity identified in #163.

## Changes

- Make it explicit that all three interfaces run in the browser.
- Preserve the existing description of Utopia as a complete, ready-to-use application.

## Scope

This PR changes one sentence in `README.md`. It is intentionally independent of the other README follow-ups so it can be reviewed and merged separately.

## Validation

- `git diff --check`
- Documentation-only change; no runtime tests required

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ The system is a Rust binary and a Postgres service. pgvector and a queue-table d
 
 | | |
 |---|---|
-| **A complete application** | A system console, a graph browser and an ontology workbench that runs in the browser. Install it and it works; there is no library to assemble first. |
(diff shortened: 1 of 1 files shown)


WaylandYang:
Thanks. #217 rewrote the Features table; the first row now says "A system console, a graph browser and an ontology workbench in one web UI", which removes the ambiguity this PR addressed. Closing as superseded.

PR #237, opened 2026-09-03
A contradiction points at an error upstream

WaylandYang:
Decision record 0017 — the full design for B2 of 0016. Contradictions between derivations and assertions become detectable (per item, capped per predicate), contradictions between derivations aggregate by the pair of rules that produced them, the Review card lays out clues and repairs instead of asking for a verdict, and a disputed fact is visible where it sits: chip in the entity panel, blocked derivations under the Derived tab, an alert colour on the graph.

Also updates the language convention in the decisions index: records from 0017 on are written in English.

No code in this PR; B2a and B2b follow it.


WaylandYang:
Superseded by #238, which carries the 0017 record together with the B2a implementation.

Issue #216, opened 2026-09-02
A failed job has no way back into the queue


WaylandYang:
When a job fails there is no way to run it again except through the object it belongs to: a document can be reprocessed, a source can be re-extracted. For anything else — `bootstrap_ontology`, `adjudicate_entities`, `sync_source`, `materialize_inferences` — a failed row in `jobs` is the end of it until whatever enqueues it comes round again.

This became worth writing down with #201. An `OutOfCredit` failure now goes to `failed` on the first attempt instead of spending three. That is right for the operator, who sees the real state seven minutes sooner, but it changes what happens to a batch:

```
500 documents queued, credit runs out halfway
before   each job retried over ~7 min; a top-up inside that window rescued the rest by itself
after    every remaining job is failed; after the top-up they have to be triggered again
```

Per document that is a click; across a batch it is the whole batch. Re-extracting the source works but re-runs the documents that already succeeded, which costs the model calls again.

What would cover it: a way to requeue failed jobs — by kind, by knowledge base, or from the alert that reported the failure. The queue already has what it needs (`status`, `kind`, `payload`, `last_error`); the handlers are idempotent, which is why orphan recovery can requeue `running` rows at startup.

Worth deciding whether the alert should carry the action. `llm.out_of_credit` is the one failure where the operator does something specific (tops up) and then wants the work to resume; an alert that can say "run those again" closes that loop where a queue page would be a new surface to build.

PR #259, opened 2026-09-03
Two live answers must not write into one store

vo1dLee3:
## The symptom

Start a question in conversation A, click over to conversation B while the answer is still streaming, and ask again. Both SSE handlers then write into one module-level store:

- B's bubble grows with A's words and tool-call steps interleaved into it — `A-2 B-0 A-3 B-1 A-4 …` — and stays on screen until you navigate away and back
- A's `finish()` pulls B's stop button while B is still streaming
- `abortRef` now points at B, so **A can no longer be stopped**: it runs to completion on its own, unbudgeted model spend with no recourse (the same credit that #195 and #216 treat as worth protecting)

Reproduced at runtime with zero patches to shipped code: the actual `liveAnswer.ts` and `api.ts` (`streamChat`/`reattachChat`, which share `consumeChatStream`) executed against a wire-identical SSE server, with the guard, send handlers, stop button and claim check copied verbatim from `Chat.tsx`. Server-side log confirms A was never aborted; a control run with no switching is clean.

## The root cause

`liveAnswer` is a singleton on a stated ground: 同时只会有一次进行中的回答. Nothing enforces that, and the rest of the file contradicts it on purpose — three places in `Chat.tsx` establish two concurrent generations as a normal, intended state:

- switching knowledge bases does not abort («换库不该杀掉另一个库里正在写的回答»)
- starting a new chat does not abort («开一场新的不等于放弃上一场»)
- the send guard is deliberately per-conversation, and its comment rejects the global-block UX («发不出消息»)

So the guard's own intent creates the state the single slot cannot hold: the second `start` replaces the store while the first stream's callbacks keep calling `patchLast` — which patches the last turn of *whatever is in the slot now*.

## What changes

**`liveAnswer` becomes a table keyed by conversation, not a slot.** Each stream gets a handle at `begin()`; every read and write goes through the handle, so A's deltas can only ever land in A's entry.

- `begin(kbId, conversationId, turns, abort)` returns a handle (`identify` / `patchLast` / `finish` / `setAbort`); a not-yet-named conversation occupies an internal token until `identify` remaps it — `kbId` disambiguates two such windows
- `entry(kbId, conversationId)` is the claim ("is the conversation I'm looking at live here"); `stop(kbId, conversationId)` is the stop button: abort + finish for exactly the entry on screen
- `begin` evicts finished entries — eviction only happens when somebody sends again, and an evicted conversation re-reads identical content from the database
- **`abortRef` is gone entirely.** It was the direct mechanism of the orphaned stream; abort now travels with its own entry
- The send guard is untouched — it already asked the right question («是「这一场」在流，不是「有一场」在流»); now the store gives it a per-conversation answer
- A side effect of the new shape: a follow-up question now builds on `liveHere?.turns ?? turns` — what is on screen — instead of the stale component state, so the previous answer no longer vanishes from the display when you ask again (fixed in a separate commit)

No UI or i18n changes; the interface looks exactly the same when only one answer streams.

## Verification

Both checks run without a backend:

- `cd web && pnpm build` green (the CI web check, includes `tsc --noEmit`)
- The repro harness, re-targeted at the new API, passes 23/23 assertions: concurrent streams stay isolated; `stop(B)` aborts only B and the server log shows it; A is then independently stoppable (no orphan, no full-length run); finished entries remain claimable (#171's display continuity) and are evicted by the next `begin`; `identify` remaps the pending entry; the reattach path (refresh-mid-stream) builds its entry from history + snapshot and writes only its own; the control case behaves as before

### DIFF
--- a/web/src/liveAnswer.ts
+++ b/web/src/liveAnswer.ts
@@ -1,4 +1,4 @@
-// 正在生成的那一次回答，活在组件之外。
+// 正在生成中的那些回答，活在组件之外。
 //
 // **切走一次就看不见了。** 流式中的 `turns` 从前是 Chat 的组件状态，而离开
 // 对话页会卸载这个组件：状态没了，那个 fetch 还在跑，回调写进的是一个已经
@@ -8,7 +8,16 @@
 // 服务端那半边（生成不随连接消失）是另一条修复；这半边解决的是**回来的时候
 // 看不看得见**。两条缺一不可：服务端保住了答案，这里保住了那条流。
 //
-// 同时只会有一次进行中的回答——所以这是个单例，不是按会话开的表。
+// **按会话键控，不是单例。** 这张表从前是一个槽位，依据是「同时只会有一次
+// 进行中的回答」。这个前提不成立，而且是被 Chat 自己否定的——换库不 abort
+// （「换库不该杀掉另一个库里正在写的回答」）、开新对话不 abort（「开一场新的
+// 不等于放弃上一场」）、发送守卫按会话收窄（明确拒绝「发不出消息」的全局封锁）。
+// 三个「不 abort」凑在一起，两场并发是常规可达的状态，而单槽装不下它：第二场
+// start 覆盖槽位，第一场的回调还在往「当前槽位的最后一条」里写，两场回答逐字
+// 交织；先结束的那场把另一场的停止按钮提前收掉，自己从此无人可停。
+//
+// 于是改成一张表：谁开场谁拿句柄，读谁写谁都有名有姓。`send` 的守卫不用改——
+// 它本来问的就是「这一场在不在流」，现在这个问题终于只关于这一场。
 import type { ChatStep, Source } from "./api";
 
 export interface Turn {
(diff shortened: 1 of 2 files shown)


WaylandYang:
Thanks for this, and for the write-up: the three "don't abort" sites and the single slot contradicting each other is exactly the kind of thing that hides until someone switches conversations mid-stream. I'm running it end to end now (two conversations streaming at once, stop on one, switch back and forth, refresh mid-stream) and will report back here.

WaylandYang:
Tested end to end against a fresh database with real documents and a real model, with the unpatched dev frontend running beside it on the same backend.

On dev the bug reproduces exactly as you described: two answers interleave word by word, with tool steps duplicated. On this branch, concurrent streams stay isolated, stopping one leaves the other running and stoppable, finished conversations stay claimable, a follow-up keeps the previous answer on screen, refresh mid-stream reattaches cleanly, and switching bases mid-stream leaves both going. `pnpm build` passes.

One adjacent thing, not from this PR: switching knowledge base inside a conversation keeps the conversation id in the URL and triggers a 404 before falling back. The old single slot happened to mask it. I'll open a separate issue.

Merging. Thank you for the careful fix and the write-up.

WaylandYang (approved):
Verified end to end (see comment above): concurrent streams isolated, stop scoped to one conversation, reattach and follow-up paths intact, web build green.

PR #264, opened 2026-09-03
A search keeps its query in the URL

vo1dLee3:
## The symptom

The search page keeps its query in component state. The URL is always bare `/kb/{kbId}/search`, no matter what you searched for.

## What a user runs into if this stays unfixed

- **Every refresh wipes the page.** Search "合同金额", results come back, press F5 — empty page, type it again. The results are derivable from the query, but the query lives nowhere the refresh can reach.
- **The back button breaks the citation loop.** Search → click a result → the doc page lands on the exact chunk (`?chunk=`, built in #144). Press Back, and the search page is blank: the loop that page exists for ends in a restart. The router restores the path fine — the query was never on it.
- **Results can't be shared or bookmarked.** Copy the URL, send it to a colleague: they get an empty search page and a description of what to type.
- **Inconsistent with everything else.** Graph (`entity/focus/at`), docs (`chunk`), library (`src`), review (`queue/item`), settings (`tab`), chat (the conversation is the route) all put "what you are looking at" in the URL. Search is the only surface where it is missing, so "the address bar can be trusted" fails exactly where the query is the purest example of it.

## What changes

The submitted query moves to the URL as its single source of truth, following the existing `validateSearch` pattern (settings tab, review queue):

- `searchRoute` declares `{ q?: string }`
- `query` is read from `useSearch`; submitting navigates with `search: { q }` — typing stays local and never touches the address bar; only submission writes it
- typing in a new query after Back/Forward follows the URL (input syncs when `q` changes); a new query resets pagination
- pagination stays local, per the graph page's own bar for "how you look" vs "what you look at" (the zoom slider never went in the URL either)

## Verification

- `cd web && pnpm build` green (the CI web check, includes `tsc --noEmit`)
- Walkthrough, no backend needed to observe: mount with `?q=x` → the search request fires on its own (results are refetched by the existing `useQuery` keyed on `q`); typing does not navigate; submitting writes the URL once; Back between two searches restores box, query and results; an empty submit is a no-op; clicking a result and pressing Back returns to a populated search page

### DIFF
--- a/web/src/pages/Search.tsx
+++ b/web/src/pages/Search.tsx
@@ -1,6 +1,6 @@
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import { useQuery } from "@tanstack/react-query";
-import { Link } from "@tanstack/react-router";
+import { Link, useNavigate, useSearch } from "@tanstack/react-router";
 import { api } from "../api";
 import { S } from "../i18n";
 import { useKb, useKbId } from "../kb";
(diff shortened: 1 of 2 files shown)


WaylandYang:
Ran it end to end on the same setup as #259. Mounting with `?q=` fetches on its own, typing leaves the address bar alone, Enter writes `q` once, Back and Forward restore the box, the query and the results, opening a result and pressing Back lands on a populated page, a refresh keeps the results, an empty submit does nothing, and switching bases from the search page lands on the other base's bare search, which is what #262 made the rule. No console errors beyond the pre-existing ones.

Merging. Thanks for another clean one.

WaylandYang (approved):
Verified end to end; see comment.

PR #265, opened 2026-09-03
An agent can read a whole document after a search hit

chetanunadkat-lang:
## Problem

`search_chunks` returns the six best sections and cuts each at 800 characters. When the answer sits past the cut, or in a section that did not rank, the agent cannot reach it. No tool and no route returns document text. Observed with a Gemini meeting-notes mail: the action items were in the tail of section 1 and in section 2, and an MCP client (Hermes Agent) tried every exposed tool and reported the knowledge base did not have them.

## Change

- New tool `get_document(document_id)` returns every current section of one document in order, capped at 24k characters with an explicit "truncated" line. It is scoped to the knowledge base in SQL (`find_in_kb`, `chunks_in_document` in utopia-store), so an id from another knowledge base reads as not found.
- Exposed over the MCP endpoint (`EXPOSED` is now five tools plus `entity_facts`). The audit row `mcp.tool_called` is unchanged.
- `search_chunks` prints `document_id` on each hit so the model can ask for the whole document. Tool descriptions and the system prompt point to that second step.
- `check_call` refuses a missing or non-uuid `document_id`, driven by `"format": "uuid"` in the schema. A made-up id would otherwise come back as "not in this knowledge base" and read as a real absence. Two tests added.
- The `search_docs` description now says it searches the product manual, not the user's documents. Name unchanged.

## Verification

- `cargo test -p utopia-server`: 120 passed on top of current `dev`.
- Live: after restart, `tools/list` shows `get_document`; a direct call on the notes document returned both sections. Through Hermes, the same question that failed before now answers with all three action items and owners, and the audit log shows `search_chunks` then `get_document`.

## Note for reviewers

The uuid gate in `check_call` applies to any required parameter that declares `format: uuid`. Only `get_document` declares it today. `entity_facts.entity_id` could get the same treatment in a follow-up.

### DIFF
--- a/crates/utopia-server/src/api/chat.rs
+++ b/crates/utopia-server/src/api/chat.rs
@@ -157,12 +157,13 @@ fn check_call(
             ),
         ));
     };
-    let required = tools
+    let function = tools
         .as_array()
         .into_iter()
         .flatten()
         .find(|t| t["function"]["name"] == name)
-        .and_then(|t| t["function"]["parameters"]["required"].as_array());
+        .map(|t| &t["function"]);
+    let required = function.and_then(|f| f["parameters"]["required"].as_array());
     for key in required.into_iter().flatten().filter_map(|k| k.as_str()) {
         // 空串与 null 都算没给：`{"query": ""}` 检索出来的东西与问题无关，
         // 而它同样会显示成一条正常的轨迹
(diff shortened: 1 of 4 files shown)


WaylandYang:
Tested end to end on a fresh base with a real model, on this branch's server.

Over MCP: `tools/list` shows `get_document`; a full read of a four-section document returns all four sections with citations, including text past the 800-character excerpt cut and in a section a search would not rank; a document id from another base comes back as not found with nothing leaked; a non-uuid id is refused; a 30k-character document is cut at 24k with the omitted count stated and the omitted sections not cited; each call leaves an `mcp.tool_called` audit row.

Through chat, the question that motivated this ("what Series C target, what P5 budget and codename, and every action item") went search_chunks → get_document on the model's own judgement, and the answer had all seven facts with citations. Without the tool the two figures sat past the excerpt cut.

`cargo test -p utopia-server`: 120 passed here too. One cosmetic note for later, not for this PR: the new step kind `document` falls through to the generic wrench icon in the chat trace; a document icon would read better.

Merging. Thank you, this was a well-scoped first contribution.

WaylandYang (approved):
Verified over MCP and through chat; see comment.

WaylandYang:
One housekeeping item I missed before merging: the commit had no DCO sign-off, and CONTRIBUTING asks for one on every commit (`git commit -s`). Could you open a small follow-up PR with a single empty commit carrying it? Something like:

```bash
git commit --allow-empty -s -m "DCO remediation for 11475a34" -m "I, chetan unadkat <your-email>, hereby add my Signed-off-by to commit 11475a34 (#265)."
```

That is all it needs. Thanks again for the contribution.

Issue #215, opened 2026-09-02
Notion needs someone with a real workspace to run its test


WaylandYang:
The Notion source ships with parsing tests only. Nobody has pointed it at a real workspace.

Unlike the other connectors there is no way to close this locally: Notion is closed-source SaaS
with no emulator, and the open-source alternatives (AppFlowy, AFFiNE) do not speak its API.

The risk is the same one that bit S3, which passed its construction tests and still broke on
every bucket with a folder in it until a live endpoint was tried.

**What would help:** create an internal integration, share one page with it, run the test.

```bash
# Settings -> My connections -> New internal integration
# then share a page with it from the page's ... menu
UTOPIA_NOTION_TEST_TOKEN=ntn_xxx cargo test -p utopia-server notion
```

A free workspace with a single page works. The token is read-only over exactly the pages you
share with it.

The parts most likely to be wrong, none of which construction tests reach:

- **Pagination.** Both `/v1/search` and `/v1/blocks/{id}/children` page at 100. A workspace with
  more than that exercises `has_more` and `next_cursor`; one with fewer never does.
- **`Notion-Version`.** Pinned to `2026-03-11`. A wrong value returns 400 saying only "missing
  version", without naming an acceptable one.
- **Database rows.** Their title property is named by the user, so the by-type lookup matters
  here in a way it does not for plain pages.
- **Rate limiting.** Roughly three requests a second. Requests are sequential for that reason,
  but nobody has watched what a workspace large enough to hit it actually does.

See `crates/utopia-server/src/notion.rs`.

jeremycapps:
Hey, I'll take this one. @WaylandYang

WaylandYang:
Thanks for picking this up — it's yours.

The command in the description is all it needs, and a free workspace with one shared page is enough. What's most useful back: whether the sync produces one document per page, whether `Notion-Version: 2026-03-11` is still accepted, and — if your workspace has more than 100 pages, or a page with more than 100 blocks — whether pagination holds. Paste any error verbatim; the client surfaces Notion's own message.

A comment with what happened closes this. No PR needed unless you find something to fix. Much appreciated.

jeremycapps:
Output from a small workspace <100 pages/blocks.
<img width="1968" height="298" alt="Image" src="https://github.com/user-attachments/assets/65defee1-27ec-4f10-a54b-4eaf30acd8ea" />

Output from a large workspace > 100 pages/blocks. I got a rate limiting error.

[...]

PR #269, opened 2026-09-03
The Notion client waits out a rate limit instead of failing the sync

WaylandYang:
Part of #215. Thanks to @jeremycapps for running the connector against a real workspace.

## Symptom

On a workspace with more than a hundred pages the sync ran for about three minutes and then died with `notion search 回了 429 Too Many Requests`. The small-workspace run in the same report passed, so the connector itself works; it is the pace that breaks.

## Cause

Requests were sequential but not paced. Notion allows roughly three requests per second; a 100 ms response followed immediately by the next request is ten. And a 429 was treated like any other failure: the sync bailed instead of waiting out the `Retry-After` header Notion sends with it.

## Change

A small `Paced` client wraps every Notion call in `notion.rs`:

- at least 350 ms between requests, which puts a 500-page sync at a little over three minutes worst case
- on 429, wait for `Retry-After` (seconds, fractional allowed; 1 s if missing or unreadable; capped at 30 s) and retry, up to five times, then surface the error
- error text is English, since it lands in the source's sync status; the previous message leaked a Chinese word into the reporter's terminal

The extraction path's backoff is not reused on purpose: it is tuned for model providers and their error taxonomy. This one is local to the connector and about twenty lines.

## Verification

- `cargo fmt`, `cargo clippy --all-targets -D warnings`, `cargo test -p utopia-server notion`: clean, 7 passed (a unit test pins the `Retry-After` parsing: header value, fractional, missing, garbage, negative, capped)
- The live test needs a real token and is skipped here. @jeremycapps, if you can run `UTOPIA_NOTION_TEST_TOKEN=... cargo test -p utopia-server notion` once more on the large workspace after this lands, that would confirm both the pacing and pagination past 100.

### DIFF
--- a/crates/utopia-server/src/notion.rs
+++ b/crates/utopia-server/src/notion.rs
@@ -19,12 +19,20 @@
 //! **`Notion-Version` 头是必填的**，而且值是日期。少了它接口直接 400，
 //! 而错误信息只说 "missing version"，不会告诉你该填哪个。
 //!
-//! **限流是每秒三次**（官方说法是「平均三次」）。所以取页面内容时是顺序而不是
-//! 并发——并发上去只会换来一串 429，而我们的重试退避是为抽取那条路设计的，
-//! 不该被摄入路径借用。
+//! **限流是每秒三次**（官方说法是「平均三次」）。顺序发请求还不够：一个 100 毫秒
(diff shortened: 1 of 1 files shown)


jeremycapps:
CleanShot 2026-09-05 at 19 30 48@2x

just reran it on the latest dev branch and all tests passed

PR #281, opened 2026-09-03
The ontology becomes visible as a schema, not only editable as forms

### LINKED ISSUE #280: The ontology should be visible as a schema, not only editable as forms
## Use case

The Ontology page (`/kb/:kbId/ontology`) lets users inspect and edit classes and properties one at a time through a master/detail form interface — pick a class on the left, see its parent, disjoint classes and attributes on the right. That works for editing a single class, but there is no way to see the ontology *as a whole*: which classes inherit from which, what object properties connect which classes, where multiple inheritance or disjointness has been declared.

Today the only way to understand the overall shape of an ontology — its class hierarchy, its domains/ranges, its relationships between classes — is to click through every class and property individually and build the picture in your head. For an ontology pack with more than a handful of classes (which is the normal case, not the exception), that does not scale.

`/graph` already gives us a great visual language for exploring the *instance* graph (entities and facts) with graphology/sigma, curved parallel edges, hover/selection, search, etc. We want the equivalent view for the *schema* itself (TBox, not ABox): classes as nodes, `subClassOf`/disjoint/object-property edges between them, navigable and searchable, the way metaphactory's schema diagrams work.

Concretely, users should be able to:
- see every class as a node, coloured/shaped as it already is in the class tree;
- see every declared parent relationship (`parents`, not only `primary_parent`) as a distinguishable `subClassOf` edge, including multiple inheritance;
- see every object property (`kind === "relation"`) as a labelled edge from its domain class(es) to its range class(es), with several parallel properties between the same two classes staying individually inspectable (e.g. `works_at`, `founded`, `owns` between Person and Organization);
- see declared class disjointness as a third, visually distinct, non-directional relationship;
- discover relations that aren't scoped to specific classes (empty domain/range) without the diagram fabricating an all-to-all fan of edges;
- click a class or a relation to see a compact read-only inspector (label, description, parents/disjoint or domain/range/axioms, usage) and jump into the existing class/property editor from there.

This is purely a new way to *look at* the ontology Utopia already stores — no new ontology data, no schema drift, no OWL/TTL export. It reuses the existing `api.ontology(kbId)` data (`EntityTypeView`, `RelationTypeView`) and the existing graphology/sigma stack from `/graph`, and lives inside the existing Ontology page as an additional primary view (alongside Classes/Properties), not a new top-level section.

kalyp-angel:
Closes #280.

## The user problem

The Ontology page lets you inspect and edit classes and properties one at a time through a master/detail form — pick a class, see its parent, disjoint classes and attributes on the right. There is no way to see the ontology *as a whole*: which classes inherit from which, what object properties connect which classes, where multiple inheritance or disjointness has been declared. For a pack with more than a handful of classes (the normal case), building that picture by clicking through every class doesn't scale.

This adds a **Schema diagram**: a graphology/sigma canvas, in the same visual language as `/graph`, that renders the ontology's classes, inheritance and semantic relations at a glance — conceptually similar to metaphactory's schema diagrams.

## Schema vs. instance — why this isn't `/graph`

`/graph` renders the ABox (entities and facts): temporal filtering, derived edges, contested edges, evidence. This is the TBox (the ontology's own structure): classes, `subClassOf`, object properties, disjointness. They share the graphology/sigma toolchain and the "chrome is neutral, colour is data" visual language, but the semantics don't overlap, so `/graph` was left untouched and the new view lives entirely inside `Ontology.tsx` as an additional primary selection (`{ kind: "schema" }` alongside class/relation/import/refine/misses), positioned above the Classes/Properties list rather than down with Import/Refine/Unmatched — it's a way of *browsing* the ontology, not an administrative action.

## Why no backend/ontology-contract change

The existing `api.ontology(kbId)` query already returns everything the diagram needs (`EntityTypeView.parents/disjoint/color/shape`, `RelationTypeView.domains/ranges/kind/…axioms`). `OntologySchemaGraph` takes that data plus two callbacks (`onEditClass`, `onEditRelation`) as props — `Ontology.tsx` remains the sole owner of the query, so there's no second copy of the ontology to go stale and nothing to invalidate separately. A pure `buildSchemaGraph(entityTypes, relationTypes)` function projects that data into a graphology graph; it's exported and has no React/Sigma dependency, so the semantics can be poked at directly (see Validation below).

## Multiple inheritance and unscoped properties

- Every declared parent in `parents` becomes its own `subClassOf` edge (child → parent, the RDF/OWL direction), not just `primary_parent` — multi-parent classes show every edge.
- Disjointness is deduplicated by sorted class-id pair before drawing, since both sides typically declare it, and rendered as a third, non-directional, restrained edge style (no arrowhead, no dashed-shader detour — a distinct hue and line weight instead).
- A relation with an empty `domains` or `ranges` is not scoped to a class, so it is **not** fanned out to every class on the canvas (which would assert something the ontology never said). It's counted in an "Unscoped properties" popover instead, where it stays selectable/editable.
- Several properties between the same class pair (e.g. `works_at` / `founded` / `owns`) get individually curved, inspectable edges rather than collapsing into one.
- Missing/stale references (a parent or domain id that no longer resolves) are silently skipped — no crash, no fabricated placeholder class.

## Reused vs. new

Reused as-is: `graphology`, `sigma`, `graphology-layout`, `graphology-layout-forceatlas2`, `@sigma/edge-curve` (`EdgeCurvedArrowProgram`), `@sigma/node-border`, and the existing `NodeSquareShellProgram`/search-dropdown/legend-chip/popover-flip conventions from `Graph.tsx` and the shared `ui` module. No new dependency was added. The layout is a synchronous circular-seed + ForceAtlas2 pass (no animated worker) — ontology packs are one to a few hundred classes, not the thousands `/graph` budgets for, so a one-shot layout keyed off `[entityTypes, relationTypes]` is simpler and doesn't restart on selection.

New, and scoped to the ontology feature only: `web/src/pages/OntologySchemaGraph.tsx`, the `{ kind: "schema" }` selection branch + rail entry in `Ontology.tsx`, and the `S.ontology.schema*` strings in `en.ts`/`zh.ts`.

## Validation

There's no existing frontend test harness (no vitest/jest in `web/package.json`, and the web CI job is `pnpm build`, i.e. `tsc --noEmit && vite build`), so per the repo's own guidance this doesn't add one just for this feature. Instead:

- `buildSchemaGraph` was exercised directly in the browser (Vite serves it as an importable ES module in dev) against synthetic fixtures covering multi-parent inheritance, self-loop relations, parallel relations, unscoped relations, bad references, and empty/single-node ontologies — all produced the expected graphs with no exceptions.
- Against a real 220-class / ~130-relation bundled ontology pack: 150 subclass edges, 34 relation edges, 22 disjoint edges (a 7-way mutually-disjoint asset-type clique correctly deduplicated to `C(7,2) = 21` edges instead of 42), 79 unscoped relations surfaced without fabricating a single cross-class edge, zero `NaN`/undefined node or edge sizes, zero console errors.
- Manually walked class selection → parents/disjoint/attributes → "Edit class" → existing `ClassForm`, and relation selection → domain/range/axioms → "Edit property" → existing `PropertyForm`; both land in the pre-existing editors with no parallel editing UI.
- `cargo build --workspace` / `cargo test --workspace` pass (no Rust files touched by this PR; `rustfmt`/`clippy` components weren't available in this sandbox to re-confirm formatting, but nothing outside `web/` changed).

I wasn't able to attach screenshots to this description from the sandboxed session this was written in — happy to add them, or a reviewer can pull the branch and open Ontology → Schema diagram directly.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


kalyp-angel:
Pushed a follow-up commit addressing feedback from a manual pass:

**Fixed a real rendering bug.** Edges silently failed to render for most relations. I'd collapsed the edge's semantic `kind` (subclass/relation/disjoint — used for color and click behaviour) and sigma's `type` attribute (which program actually draws the edge) into one field. `EdgeCurvedArrowProgram` was registered as the default for anything not explicitly "subclass" or "disjoint", but the vast majority of relation edges (no parallel siblings) have `curvature === 0`, and a curved-arrow program at zero curvature wasn't drawing visibly. Fix: `kind` (semantics) and `type` (render dispatch) are separate attributes again; a relation edge now gets `type: "arrow"` (straight, `EdgeArrowProgram`) when it has no parallel siblings and `type: "curvedArrow"` only when it actually needs to curve. Verified against the 220-class pack: 184 straight + 22 line (disjoint) + 0 curved (no parallel relations in that pack) — 206/206 edges resolve to a real program, 0 fall through to "other".

**Editing now happens in place.** Selecting a class or relation on the canvas no longer opens a custom read-only summary with a link out — it embeds the actual `ClassForm` + `AttributesCard` / `PropertyForm` from the main Ontology page directly in the side panel (exported from `Ontology.tsx`, not reimplemented). Same mutations, same validation, same `onDone`/`onError` wiring — there's still exactly one editing implementation, it's just now reachable from both places.

**Visual parity with `/graph`.** Extracted the generic (non-instance-specific) parts of `/graph`'s node rendering — the four-layer shell color recipe, the pill-label/hover-card canvas draw functions, and the world-coordinate grid background — into `web/src/pages/graphVisuals.ts`, and `Graph.tsx` now imports from there instead of defining them locally (net reduction in `Graph.tsx`, not a copy). The schema diagram uses the same recipe, so a class node and an entity node in `/graph` read as drawn by the same engine. Edge treatment (color-by-kind, no derived/contested/temporal concepts) stays schema-specific, per the original scope.

Re-verified: `pnpm build` clean, manual pass over both `/graph` (pixel-identical before/after the extraction) and the schema diagram (edges visible, inline forms save correctly, no console errors).

WaylandYang:
Thanks for this, it's careful work and it answers #280. I pulled the branch and ran it against a live backend: on a six-class base everything does what it says, and /graph is untouched by the visuals refactor.

Having used it on the two kinds of base we actually have, though, I want to suggest a different shape for the same idea, and I'd rather say it now than after another round.

A whole-ontology view rarely has a readable state here. A new base installs schema.org by default, 1010 classes: the diagram blocks the main thread for about 1.2 s, appears after five, and what appears is one dense disc with every label on top of every other. A base that grew its ontology from documents is the opposite: its properties have no domain or range, so there is nothing to draw, six dots and an "Unscoped properties" popover. Between those two, the overview is either too much or empty, and the "few hundred classes" middle it is designed for is not a base anyone here has.

What always reads is the neighbourhood of one class. So: instead of a top-level "Schema diagram" entry, put a small panel inside the existing class editor (and the property editor) that draws just what is adjacent to the thing being edited: its parents and children, the classes it is disjoint with, the classes its object properties connect it to, with the same edge styles you already have. Clicking a neighbour navigates to it, so the diagram becomes the way to walk the ontology one hop at a time, and it stays legible at 6 classes and at 1016. `buildSchemaGraph`, the sigma setup and `graphVisuals.ts` all carry over; what goes is the whole-graph layout and the rail entry. A "Show all" action can come later if a real base ever wants it, but it should not be the default.

If you're up for that, the branch is most of the way there and I'll merge it once the panel is in. If you'd rather not reshape it, say so and we can talk about which parts to land as they are.

kalyp-angel:
Found the actual bug behind "editing relationships isn't working." Clicking a relation edge on the canvas wasn't registering at all — search-based selection worked, but the canvas click didn't.

Root cause: sigma's edge hit-testing (`getEdgeAtPoint`) picks against the same WebGL geometry the edge is rendered with — there's no separate "click tolerance" padding. The default `minEdgeThickness` is 1.7px, and relation edges were drawn at `size: 1.3`, so the actual clickable band was only a couple of pixels wide. Verified this precisely: swept `getEdgeAtPoint` in a grid around a computed edge midpoint and the hit region was a thin, off-center sliver — real mouse clicks would routinely miss it and fall through to `clickStage`.

Fix: relation edges are drawn thicker (`size: 1.3 → 2`) and `minEdgeThickness: 3` is now set globally. This serves two things at once — the spec already wanted relation edges "visually more prominent than structural helper information," and a thicker line is also a bigger, actually-clickable target. Re-verified with the same sweep method: the hit region went from a handful of scattered points to a comfortable ±6px band, and a full click→`clickEdge`→panel-opens round trip now succeeds reliably.

kalyp-angel:
Pushed a larger follow-up addressing another round of feedback: creating/removing relationships from the diagram, one unified editing surface instead of two, and a post-edit consistency nudge.

**One workspace, not two.** Previously the schema diagram was a separate mode from the classic master/detail form — selecting a class from the left rail would kick you out of the diagram back to the old full-page form. Now `class` / `relation` / `new-class` / `new-relation` / `schema` / no-selection all share one workspace: the diagram is the permanent backdrop, and the *actual* `ClassForm` / `PropertyForm` / `AttributesCard` (exported from `Ontology.tsx`, not reimplemented) dock on the right when something's selected. Left-rail click, canvas click, and the diagram's own search box all funnel into the same `sel` state, so there's exactly one editing surface regardless of entry point. `OntologySchemaGraph.tsx` shed ~160 lines in the process — it no longer needs to know what a form looks like, only what's selected (`selected`/`onSelect` are now controlled props from `Ontology.tsx`).

**Docked panel is solid, not glass.** The floating panel previously used the same `glass`/`glass-strong` (blurred, translucent) treatment as everything else in the app — reasonable over a static background, but distracting over an actively-animated WebGL canvas showing through the blur. Switched to `.u-modal-panel` (~97% opaque), which already exists in this codebase for exactly this reason — its own comment describes solving the identical problem for confirmation dialogs ("reading a decision against legible text bleeding through underneath is a direct distraction").

**Create/remove relationships from the diagram.** Deleting already worked (`PropertyForm` has always had a delete button). Creating didn't — there was no entry point. Selecting a class now shows an "Add relationship…" button that opens the (also unified) new-property form with the domain pre-filled to that class, so a previously-unconnected node isn't a dead end anymore. (`PropertyForm` gained one new optional prop, `initialDomain`, for this.)

**Consistency check after every ontology edit.** `ClassForm`/`PropertyForm`/`AttributesCard` now funnel their `onDone`/`onChanged` through one `afterOntologyChange` that both refreshes the query and calls the existing `api.runConsistencyCheck` (the same synchronous, non-destructive endpoint Review's own "Run check" button uses) — if it surfaces new ontology defects, a toast links straight to Review. I deliberately did **not** wire this to `api.rebuildGraph` — that's a different, admin-gated, typed-confirmation-required operation that clears and re-derives the entire graph, and auto-triggering something destructive after a routine field edit would be the wrong call. Happy to add an explicit link to Library's rebuild flow if a full re-extraction pass is actually what's wanted here, separately from this safe check.

**What I deliberately didn't build:** class-merge / "these two classes look like duplicates" suggestions. I checked — entity *instances* can already be merged (Review's duplicate cards, `api.mergeEntities`), but there's no equivalent for ontology *classes* anywhere in the backend, and building one means new surface area (reassigning `relation_type_domains/ranges`, disjoint declarations, attributes, and instance types across the merge). That's real ontology-contract territory, which this repo's CONTRIBUTING.md is explicit needs a discussion + ADR before code, not something to fold into this PR unannounced. Flagging it back to the requester rather than guessing at the design.

Verified live: rail-click keeps the diagram visible with the real form docked (screenshot-confirmed opaque, not see-through), "Add relationship…" correctly pre-fills domain, close button clears the panel back to the bare diagram, `/graph` unaffected, no console errors. The one thing I couldn't verify end-to-end in this sandbox is a *successful* save + consistency-check round trip — my test account gets a 403 from this KB for any ontology write, which I confirmed is a permissions issue (same 403 on a direct `runConsistencyCheck` call, unrelated to this change) rather than anything in the new code; the request itself reaches the right endpoint with the right payload.

kalyp-angel:
Pushed one more piece: a selected class now shows what it's already connected to, and lets you connect it to something else using a relationship that already exists (rather than only ever creating a new one).

**\`Relationships\` card**, docked below the class form, split into:
- **From this class** — every relation where this class is a domain, with the range side shown (e.g. \`performs → ServiceRole, Role, Task\`).
- **To this class** — every relation where it's a range, same idea in reverse. A relation with this class on both sides (a self-loop, e.g. \`reports_to: Person→Person\`) correctly shows in both.
- Each row is clickable and opens that relation's real \`PropertyForm\`, same as everywhere else in this PR.

**Connect using an existing relationship.** This is the part that didn't exist in any form before: pick a relation that already exists elsewhere in the ontology (e.g. \`works_at\`, already scoped to \`Person → Organization\`), choose whether this class joins it as Subject or Object, hit Connect — it does an \`updateRelationType\` that appends this class's id to the relevant side (deduped, additive only — never removes an existing domain/range), then jumps to the now-updated relation so the change is visible immediately. This is the actual ontology-modeling move the request was asking for: reusing \`works_at\` across \`Person\` and \`Contractor\` rather than growing a new near-duplicate relation per class pair.

The existing \"create a brand new relationship\" affordance is still there too, renamed \`New relationship…\`, now living at the bottom of this same card instead of floating on its own.

Verified live against the bundled pack: the card correctly lists a 3-relationship class split into from/to, the existing-relationship search picks up real relations with key hints, and clicking a listed relationship correctly opens its form. I could not complete a full successful Connect round-trip in this sandbox — my test account gets a 403 on any ontology write to this particular KB (confirmed separately, unrelated to this change) — but the request reaches \`PATCH .../relation-types/:id\` with the right merged domains/ranges payload, and the pre-existing \`PropertyForm\` save path this reuses is unchanged.

kalyp-angel:
Pushed node dragging.

Classes can now be dragged to a new spot; the position sticks and stays put — there's no live force simulation running (the layout is a one-shot pass, not the animated worker \`/graph\` uses), so a dragged node doesn't get pulled back. Hover shows a grab cursor, drags farther than 4px so it doesn't fight with click-to-select, and dragging correctly suppresses selection (confirmed: a real click still opens the docked form, a drag doesn't).

The part that took more care: **manually-placed positions survive an ontology edit.** Every save (create/update/delete a class, connect a relationship, anything that goes through the `afterOntologyChange` path from the last round) refetches `entityTypes`/`relationTypes`, which — since the diagram rebuilds its graph whenever those arrays get a new reference — would otherwise wipe out any manual arranging back to the auto layout on literally every edit. Given how central editing now is to this view, that would've made dragging nearly useless in practice. Fixed by keeping dragged positions in a ref that survives graph rebuilds (session-scoped, not written to the backend — same as how `/graph`'s own drag positions aren't persisted either) and re-applying them after each rebuild's automatic layout runs. Verified this specific mechanic directly since I couldn't drive a full live edit in this sandbox (same pre-existing account-permission wall as the last two updates): built a graph, dragged a node, rebuilt with a class added (simulating what a save does), and confirmed the dragged node kept its manual position while the new class still got auto-placed.

Also verified live: dragging visibly moves a node, "Fit view" correctly re-frames to include the new position (had to explicitly clear sigma's custom bounding box on drag-end for this — otherwise fit-view would keep using a stale pre-drag frame), and a plain click still opens the class form as before.

WaylandYang:
Thank you for this — it's a substantial piece of work, and it shows. The TBox/ABox separation, the decision not to fan out unscoped properties, and the note on why the panel stays near-opaque instead of glass over a moving canvas are all the right calls, and the writeup made the review easy.

I have it running locally and I'm making some small adjustments on top. They're all panel chrome, nothing about the diagram semantics: the detail panel gets the same shell as the entity panel on `/graph` — a fixed header carrying the class name, usage and the close control, with only the body scrolling — instead of the separate close button above a fully scrolling column. Your near-opaque surface stays.

I'll push those to this branch rather than sending you a list of nits, so do a `git pull --rebase` before your next commit. Say the word if you'd rather have them as review suggestions and I'll move them over.

Two housekeeping notes: the branch is behind `dev`, where the search-icon offset in the class form is already fixed (#288), so a merge from `dev` will take care of that one; and the CI checks still need a maintainer to approve the workflow run for a fork PR — we'll get that going.

Issue #270, opened 2026-09-03
Two namesakes with identical profiles attach a later mention by candidate order


WaylandYang:
Follow-on from #221, which covers two namesakes inside one document. This one is about what happens afterwards.

## Symptom

Once a base holds two entities with the same name, say "Zhang Wei" A (Platform Engineering) and B (Finance), a later document that mentions "Zhang Wei" without any handle is attributed by profile similarity. If both came out of the same chunk, their profile vectors are identical, so the new mention scores the same against both. `resolve_mention` keeps the first candidate it met (`sim > s`, strict), and once that score clears `SIM_ATTACH` it attaches silently. No review pair is filed. Which entity gets the fact depends on the order the candidates came back from the database.

A meeting note saying "Zhang Wei reported the quarter's budget" is plainly the Finance one, but the clue is in the facts already on B (department, title), not in a whole-chunk embedding, so it is never used.

## Where

`crates/utopia-store/src/resolution.rs`, `resolve_mention`: the scoring loop and the `sim >= SIM_ATTACH` branch. The profiles start identical because `update_profile` builds a centroid from chunk embeddings, and both namesakes were seeded from the same chunk.

## Suggested direction

Two candidates with the same name whose scores are within a small margin of each other should not attach to either. Options, not mutually exclusive:

- treat a near-tie among namesakes as grey zone regardless of the absolute score: attach to neither, create the entity, file review pairs against both (stage `human`, the adjudicator would see two identical records and flip the same coin)
- let disambiguating facts take part in the tie-break: compare the mention's chunk text against each candidate's `works_for` / `job_title` objects, which is what a person would do

Do not touch `SIM_ATTACH` / `SIM_NEW`; this is about ties, not thresholds.

## Test

Database-backed, in `crates/utopia-store/tests` behind `utopia_store::test_db::url()`: seed two same-name entities with identical profile vectors and different `works_for` facts, resolve a mention with a context vector similar to both, and assert that no ATTACH happens and two review pairs exist.

Blocked on #221 landing, since that PR is what makes two namesakes from one document exist in the first place.

WaylandYang:
[...]

PR #296, opened 2026-09-04
Preserve same-name entities with response-local extraction handles

### LINKED ISSUE #221: Two people with the same name in one chunk become one entity
**Repro:** a CSV roster with two rows named "Zhang Wei" (Platform Engineering / Shanghai / Senior Engineer, and Finance / Beijing / Financial Controller), plus board minutes that mention the Finance one. After extraction there is a single "Zhang Wei" entity carrying `job_title Senior Engineer` and `job_title Financial Controller`, `works_for Platform Engineering department` and `works_for Finance department`. Nothing lands in Review > Duplicates.

**Why (as far as I can see):** both mentions come from the same chunk, so they share the chunk profile vector; stage 2 (profile similarity) scores them as identical and attaches, and stage 3 (adjudicator) never sees the pair. "Same name is not the same person" only holds across chunks today.

**Suggestion:** when two mentions with the same name appear in one chunk with different attributes or departments, treat them as distinct candidates, or at least route the pair to the adjudicator / review queue instead of attaching on an identical embedding.

Found in the end-to-end pass on dev @ ce75740.

Recoletas:
Closes #221.

Adds response-local entity handles and fact refs so same-name entities stay distinct through extraction and fact binding.

Namesakes get separate provisional UUIDs and human-only review. Later ambiguous bare-name mentions use a document-scoped provisional entity instead of guessing between existing namesakes. Legacy no-handle behavior, cross-document thresholds, and the existing run-end disambiguator refresh remain unchanged. No migration is required.

### Validation

- `cargo test --workspace`
- PostgreSQL-backed namesake/review tests

### Benchmark

Same MiniMax-M3 + embo-01 setup, fresh KB per run:

| Corpus | upstream/dev | this branch |
| --- | --- | --- |
| pharma | 26 entities / 22 facts / 1 absent | 31 / 24 / 4 absent |
| tech | 25 entities / 24 facts / 2 absent | 27 / 23 / 1 absent |

A repeated pharma run gave `31 / 28 / 2 absent` upstream vs `34 / 29 / 1 absent` on this branch. `wrongly changed` and extraction drops were 0 in all runs; the earlier absent increase was not reproduced, so no regression or improvement is claimed.

Cross-document namesake ties remain tracked in #270.

Touches decision record 0006.


WaylandYang:
Tested end to end on a fresh base with the schema-org pack, three small documents.

One document introducing two John Smiths (a CEO at Acme Robotics, a senior nurse at St Mary's) now yields two Person entities with their facts on the right one: works_for → St Mary's Hospital (2008) and job_title → senior nurse on one, works_for → Acme Robotics (2015) and job_title → chief executive (2019) on the other, first_met linking the two. One human-stage namesake review appears in Duplicates and nothing wakes the adjudicator. A second document with one person mentioned several times still resolves to one entity. CI is green.

Three things I noticed, none of them for this PR:

- Both John Smiths show the disambiguator "Person", because `refresh_disambiguators` only reads works_at / part_of / located_in / leads and schema-org uses works_for. Pre-existing; this PR just makes it visible. Filed as #299.
- The review card renders the namesake sentinel score as "100% context similarity". The card should read the reason instead. Filed as #300.
- A third document with a bare "John Smith" that is clearly the nurse attached silently to the CEO. That is the cross-document case you left to #270; I added it there as a reproduction.

Merging. Thanks for the careful work on this, and for the benchmark numbers.

PR #310, opened 2026-09-04
Do not present namesake reviews as 100% context similarity

### LINKED ISSUE #300: The review card says why a pair is queued
A namesake review created by extraction (#296) carries score 1.0 as a sentinel. The Duplicates card renders that as "100% context similarity", which reads as strong evidence the two are the same, while the reason field says the opposite: the extractor saw two people with one name in one document.

The card should render from `reason` when it is namesake ("Two people with this name in one document") and show the similarity figure only for profile-based reviews.

Floating-Y:
## What changed

- Translate the `namesake` review reason into user-facing English and Chinese text.
- Hide the context similarity percentage for namesake reviews.
- Keep similarity visible for profile-based reviews.

## Why

Namesake reviews use `1.0` as a sentinel score. Rendering it as “100% context similarity” suggests strong evidence that the two entities are identical, even though the review was queued because the extractor found two people with the same name in one document.

## Testing

- `node node_modules/typescript/bin/tsc --noEmit`
- `node node_modules/vite/bin/vite.js build`

Closes #300


WaylandYang:
Reviewed and verified against the backend: `create_namesake_reviews` writes the bare code `namesake` with the sentinel score, so the `split("|", 1)` guard also survives the two suffix forms (`|redirected` from a merge redirect, `code|detail` from an escalation), and the new key fixes a second thing on the way — before it, `escalationText` found no wording and rendered the raw `namesake` to the user.

One change pushed on top: the reason path is type-agnostic — `type_id` is whatever extraction assigned, so two same-name organizations or projects reach the same card. The copy now says "Two entities with this name in one document" (and the Chinese to match) rather than "people". The issue's own wording was the one that presumed people; that was our text, not yours.

Merging once it is up to date with dev.

Issue #307, opened 2026-09-04
The graph slider cannot rewind the recording axis


WaylandYang:
Implements [0019](docs/decisions/0019-the-second-clock-can-be-rewound.md).

The entity panel's History view already rewinds the recording axis for one entity. What is missing is the whole graph: `edges_among(at)` carries an unconditional `invalidated_at IS NULL`, which appears 26 times in `graph.rs` and 54 times across the store, so a fact corrected in March is absent at every slider position.

0019 settles the design. One predicate, `held_at(T)`, replaces the hard-coded filter on every read; write paths keep it, because a correction is never made as of March. `at` (world) and `as_of` (record) stay two parameters all the way out to the API, since folding them into one control answers one question with the other and both look plausible on screen.

The risk named in the record is a forgotten WHERE — the failure mode 0009 was bitten by, where a guard reads correctly in Rust and runs silently empty in SQL. So the predicate goes in one place per table, and a database-backed test asserts both directions: a retracted fact absent at `as_of = now`, present before its `invalidated_at`, and a fact recorded after T absent at T.

Read-only. No new column, no migration.

Issue #277, opened 2026-09-03
A business rule over attribute facts should classify an entity, with premises


WaylandYang:
Raised by a user describing a mud-logging workflow: a well counts as gas-bearing when its total-hydrocarbon reading is above a threshold **and** the interpretation category is "gas anomaly" or "post-effect gas anomaly". Both values sit in a paragraph of the mud-logging report (an Excel sheet) or in the DDMS.

## What works today

- The two inputs can be captured. Define `Well` with two attributes (`total_hydrocarbon`, number, %; `interpretation_category`, text) and extraction turns "全烃 12.3%，解释为气测异常" into dated attribute facts with the source passage attached (attribute facts: `relation_types.kind = 'attribute'`, values in `facts.object_value`).
- If the DDMS is Postgres it can be mounted and the question asked in chat; the agent writes the SQL. Other engines are on the roadmap.
- For a handful of wells, asking the question in chat works: the agent reads the attribute facts and compares. It does not scale past tool-output limits and it is not a stored result.

## What is missing

The rule itself has nowhere to live. The reasoning engine compiles ontology axioms only (transitive, symmetric, inverse, sub-property; 0002 rules out a user-defined DSL), and derivation runs over entity–entity facts: literal-valued attributes do not take part (`derived_facts.object_id` is NOT NULL). So "value > threshold" cannot be expressed, and "this well is gas-bearing" cannot be a derived fact with premises.

## Shape of the feature

A rule layer over attribute facts, declared by people, evaluated like derivations:

- **Declaration**: a rule names a class or predicate to conclude, and a conjunction of conditions over one entity's attribute facts: numeric comparisons (`>`, `>=`, `<`, `<=`, ranges), category membership (`in {…}`), presence. Authored in the ontology workbench or as a file; never proposed by the model (same line as axioms: the reasoning engine's criteria are written by a person).
- **Conclusion**: a derived typing (`Well` → `GasBearingWell`) or a derived attribute fact (`gas_potential = good`), stored with the rule and premise facts like any other derivation, drawn as derived on the graph, never overriding an asserted fact.
- **Time**: validity is the intersection of the premise intervals, so a well that was gas-bearing on the 2023 reading and not on the 2025 one gets two intervals, not a flip.
- **Maintenance**: change the threshold, rerun; premises retracted, conclusion invalidated; same cadence and switches as materialised inference.
- **Explanation**: the entity panel shows the rule and the two facts, each pointing back to its passage.

## Not in scope

[...]

Issue #242, opened 2026-09-03
Snowflake needs someone with a real cluster to run its test


WaylandYang:
Snowflake support landed in #239 with wiremock replays of the protocol, but it has never run against a Snowflake account with a programmatic access token or OAuth token. If you have one, please register it under System settings > Data sources as

```
snowflake://:TOKEN@account.snowflakecomputing.com/DB/SCHEMA?warehouse=WH
```

then Test, mount it on a base, refresh the schema, and ask Chat a question that needs a query. What we need back: whether Test and the schema sync succeed, what the schema document looks like (table count, comments present or not), and one query_data round trip — the SQL the model wrote and what came back. Anything that fails, paste the error text; the engine surfaces the service's own message.

See docs/decisions/0017 for what is deliberately not supported yet (key-pair JWT for Snowflake, Kerberos / OAuth2 for Trino).

Issue #241, opened 2026-09-03
Databricks needs someone with a real cluster to run its test


WaylandYang:
Databricks support landed in #239 with wiremock replays of the protocol, but it has never run against a Databricks workspace with a running SQL warehouse. If you have one, please register it under System settings > Data sources as

```
databricks://:TOKEN@host/sql/1.0/warehouses/ID?catalog=main
```

then Test, mount it on a base, refresh the schema, and ask Chat a question that needs a query. What we need back: whether Test and the schema sync succeed, what the schema document looks like (table count, comments present or not), and one query_data round trip — the SQL the model wrote and what came back. Anything that fails, paste the error text; the engine surfaces the service's own message.

See docs/decisions/0017 for what is deliberately not supported yet (key-pair JWT for Snowflake, Kerberos / OAuth2 for Trino).

PR #332, opened 2026-09-04
A rule reads attributes and concludes a type (0021)

WaylandYang:
Record only — no code. Settles the shape of #277 before any of it is built, and corrects two status lines that had drifted.

## 0020 · A rule reads attributes and concludes a type

A mud-logging user wants a well marked gas-bearing when its total-hydrocarbon reading clears a threshold **and** its interpretation category says so. Both values are already dated attribute facts with their passage attached; the rule that reads them has nowhere to live. The reasoning engine compiles ontology axioms only, and `derived_facts.object_id` / `predicate_id` are both `NOT NULL`, so a derivation is always an entity–entity edge — a literal comparison cannot be expressed and neither conclusion #277 names (a derived typing, a derived attribute) can be stored.

Four decisions:

1. **`derived_facts` widens to match `facts`** — nullable `object_id`, an `object_value`, and the same `object_id IS NOT NULL OR object_value IS NOT NULL` check the facts table already carries. The proof chain, invalidation, the materialisation pass and the derived-over-asserted precedence then work on a literal-valued conclusion unchanged, because they key on the row's table, not on its object being an entity. The alternative — a `derived_typings` table beside a separate widening — adds two shapes where the ground already had one general one, which is the seam 0016 exists to stop opening.
2. **A derived typing is a derived attribute fact on a builtin `is_a`**, naming the class by IRI. It does not write `entities.type_id`: that column holds the asserted type and a rule conclusion sits below it (0015). With decision 1 in hand this needs no second concept.
3. **Rules live in `attribute_rules`, authored by a person, never proposed by the model** — the governance line 0002 drew for axioms. Conditions are structured `(attribute_predicate, op, operand)` over one entity's own attributes; no arithmetic across entities, no aggregation, no general rule language.
4. **Validity is the intersection of the premise intervals** — a 2023 reading that fires and a 2025 one that does not give two intervals, never a row that flips. This is the only new temporal logic in the cut.

The record also explains why the evaluator is a second pass rather than an extension of `derive()` (attribute facts never enter the edge walk, and restricting conclusions to types and attributes keeps the two passes free of a cycle by construction), phases the build in five steps, and records two dead ends and four open questions.

## Two status lines corrected

The convention says the PR that implements a record updates its status line. Two did not:

- **0018** claimed **none** of the lakehouse engines had run against a real cluster. Trino has, since #327, and #240 is closed. Databricks and Snowflake still want one (#241, #242).
- **0019** claimed nothing was built. #317 landed the read paths and the API; the graph control is the remaining cut (#307).


WaylandYang:
Renumbered to **0021** in 2844803 — 0020 was taken by "An auditor reads it without us", already on `dev`.

That record never reached the index, so the same commit adds its row. Renumbering without it would have left the index running 0019 → 0021, and the conventions say four digits, increasing, no gaps.

The branch is 7 commits behind `dev`, so the new 0020 row links to a file that only exists after a rebase or merge. No force push — the history is untouched, just `git pull`.

PR #480, opened 2026-09-08
Classes, properties, attributes and knowledge edit in dialogs; the panel shows

WaylandYang:
The right-hand panels on Ontology and Graph now only show; every change opens a dedicated dialog.

## What changes

- **`FormDialog`** (`ui/dialog.tsx`): one shell for edit dialogs — title, optional description, the body in a `<form>` (Enter in a field saves; the selects already swallow their own Enter), Cancel and Save at the bottom right, an optional Delete on its own at the bottom left. Always mounted, conditionally rendered, like `DangerConfirm`.
- **Ontology** (`pages/ontologyDialogs.tsx`): `ClassDialog`, `PropertyDialog` (wide), `AttributeDialog`, `ConnectDialog`. The three inline forms (`ClassForm`, `PropertyForm`, `AttributeForm`) and the connect widget leave `Ontology.tsx`; the fields are unchanged. The docked panel gets an `actions` slot (a pencil beside the close key) and shows read-only definitions: parent, disjoint, description for a class; subject, object, temporal semantics, the axioms that are on, inverse, super-property, description for a property. Attribute rows open the dialog; "New attribute", "Connect an existing relationship…", "New relationship…" and the rail's "New class / New property" open dialogs too; "New sub-class" sits under the definition. `Sel` loses its `new-class` / `new-relation` states — creation no longer docks a form.
- **Graph** (`pages/graphDialogs.tsx`): `EntityDialog` (name + type, replacing the inline block under the panel header) and `FactTimeDialog` (start, end as three radios, why — replacing the editor that unfolded inside a timeline row).
- `DESIGN.md` rule 6 gains the sentence: a floating panel shows; it does not edit.
- Strings: `ontology.edit/editClass/editProperty/editAttribute/connectTitle/connectOpen/noDescription/axiomsNone`, `graph.editTitle`, `newSubClass` retitled (it carries a Plus icon now), en and zh.

## Verified

- `tsc`, style guard clean.
- In the browser on this branch: the class panel shows the definition with the pencil → Edit class dialog (Delete left, Cancel/Save right); Attributes tab → New attribute dialog; Relations tab → the two entry buttons → Connect dialog; a property → read-only definition → Edit property dialog; on Graph, the entity pencil → Edit entity dialog with the type pre-selected; Timeline row pencil → Correct the interval dialog. No failed requests.


WaylandYang:
Round 2, from looking at 5198:

- **Relations**: "From this class" / "To this class" fold (a Row with the chevron in its icon slot, body indented 24 so the rows' arrows sit under the group's arrow); the two entries below are rows, not buttons.
- **Definition**: "New sub-class" is a row like the rail's "New class", no divider above it; "New attribute" the same.
- **Hover card on both canvases**: the Tooltip's skin exactly — `.u-pop`, 4 px radius, 8/4 padding, 12 px text; no glow, light shadow.
- **Schema diagram**: parallel layout now covers every edge kind, not only relations — an inheritance edge, a disjoint edge and a relation on the same pair of classes (Unit / Organization in the demo base) each get their own arc; a curved no-arrow program is registered for disjoint.

WaylandYang:
Round 3 — the entity panel's Relations, after the question "how do others do it":

- **One list instead of two tabs.** Timeline folds into Relations: two sections, **From ⟨entity⟩** / **To ⟨entity⟩** (collapsible, the same chevron-in-a-Row head as the Ontology panel), predicate groups inside with no direction arrows — the section says who the subject is. Within a group rows sort by start; facts that no longer hold fold under "N past" at the group's tail, dimmed with their end. Tabs are now Relations | History | Derived.
- **Rows are rows.** No leading chevron; the name is the link (navigates), chips follow, the interval sits on the right. Hover reveals "N sources" (opens the evidence under the row — Wikidata's references pattern) and the pencil (interval dialog), so the Timeline's one unique feature travels with the row.
- Strings: `graph.fromEntity/toEntity/past/sources` en + zh.

WaylandYang:
Round 4: the entity panel's rows follow the Ontology page's Relations exactly — no predicate sub-groups. Under **From ⟨entity⟩ / To ⟨entity⟩** each fact is one row: the direction arrow in the icon slot, the relation name on the left (with disputed / % / unconfirmed after it), the interval and the other entity's name on the right; the whole row navigates. Hover still reveals "N sources" and the pencil; facts that no longer hold fold under "N past" at the section's tail.

Issue #488, opened 2026-09-08
A rule can conclude a computed value, not only a constant


WaylandYang:
Today a rule's conclusion is a **constant**. `attribute_rules.conclude_value` is a JSONB literal, and `Conclusion::Attribute { predicate, value }` copies it straight into `derived_facts.object_value`. So `grade = A` can be written; `margin = revenue − cost` cannot. A condition's operand is a constant too — you can ask `revenue > 100`, not `revenue − cost > 0`.

There are real cases for computing. This issue is about letting an operand — on either side of a rule — be an expression over the subject's own attributes.

## The shape: a picked tree, not a typed string

A class/attribute picker plus an operator picker, storing a small structured tree. **Not** a formula text box.

A text box is not wrong because it would break the display — a string parsed at save time into the same tree keeps every property we care about (the rule still renders as a sentence, still draws as one edge, two rules still diff). It is wrong for two other reasons:

- **Errors move from unfillable to undiscovered.** A picker can only produce a well-formed expression against predicates that exist. A box lets you save `revenue − cost` on a class that has no `cost`, and the symptom is a conclusion that quietly does not appear at the next materialisation.
- **A grammar grows on request.** Once there is a box, `if`, rounding, `coalesce` and `sum(related.x)` are each one line of asking. The first few are merely work. **Aggregation actually breaks the model**: the premises of a `sum` are a set whose membership changes, and every derived fact here points at the readings that made it true and retires when they change (the same reason [0026](../blob/dev/docs/decisions/0026-a-rule-may-say-or-once.md) refused an arbitrary and/or tree).

The concession worth writing down: a picker is right for one operator and wrong for four. `(revenue − cost) / revenue` as nested dropdowns is worse UI than typing it. If the shapes people actually want turn out to be that deep, a box that parses into the same tree is the answer, and nothing above forbids it.

Storage: a JSONB node — `{"attr": <predicate_id>}` | `{"const": <n>}` | `{"op": "add"|"sub"|"mul"|"div", "l": <node>, "r": <node>}` — with a depth cap. The same node type serves a condition operand and a conclusion value.

## What it changes

- **Premises and interval: nothing new.** An expression reads N attribute facts; all N are premises and the conclusion holds on their intersection, which is what `validity()` already computes for a conjunction of conditions. Retirement works unchanged.
[...]

PR #493, opened 2026-09-08
A rule computes what it concludes, written down

WaylandYang:
Cut 1 of #488. `docs/decisions/0032-a-rule-computes-what-it-concludes.md`, plus its index row. No code — this is the argument, so it can be disagreed with before anything is built.

A rule's conclusion is a constant. `attribute_rules.conclude_value` is a JSONB literal and a condition compares an attribute against a literal, so `grade = A` can be written and `margin = revenue − cost` cannot. The number gets computed in a spreadsheet and typed back in as an asserted fact — losing the two readings it stood on, and going stale the moment either changes.

**An operand becomes an expression tree**, on both sides: `{"attr": id} | {"const": n} | {"op": …, "l": …, "r": …}`. A bare const is what every rule stores today, so nothing migrates. The premises and the interval come out right with no new machinery — an expression reads N facts, all N are its premises, and the value holds on their intersection, which is what `validity()` already computes.

**Picked, not typed** — and the record is explicit that the obvious reason is the wrong one. A string parsed at save time into the same tree keeps everything that matters: the sentence still renders, the edge still draws, two rules still diff. The real reasons are that a picker cannot compose an expression over a predicate that does not exist on the class (a box lets you save it, and the symptom is a conclusion that quietly stops appearing), and that a text box grows a grammar on request. With the concession written down: a picker is right for one operator and wrong for four.

**A value may be reached across a relation.** `well → field → depth` has definite premises — the relation fact plus the attribute fact — and an interval that is their intersection, so a well that sat in that field only 2019–2022 contributes that value only then. A to-many relation is not a problem either: two `field` facts give two candidate values, each with its own premises, which is the same cartesian product the evaluator already walks. The implementation cost is named honestly: the evaluator reads only the literal channel today, so a path needs the entity–entity channel too, and that is the bulk of the work rather than the arithmetic.

**Aggregation is refused**, and the line is not "one value versus many". A `sum`'s premise would have to include *and there are no others*, which is not a fact this base can hold — open world, extraction lags documents, documents lag the world. A conclusion resting on an absence has nothing that can retire it when the absence ends, the same reason 0029 refused "no such attribute at all".

Three costs it does not hide: **row count** (a computed conclusion is one row per combination of readings, so `MAX_COMBOS` starts bounding output rather than premise expansion, and wants measuring against a real base first), **units and datatypes** (`relation_types` carries them and nothing compares them — `revenue (USD) − cost (EUR)` has to be refused where it is written), and **absence and division by zero** (no conclusion, reported the way `capped` is, because "not computed" must not look like "not matched").


WaylandYang:
Corrected the aggregation section. The first draft said a `sum`'s conclusion has "nothing that can retire it" — that is wrong, and the mistake is worth naming: retirement here is a **full recompute plus diff**, so a fourth well simply produces 400, the 300 falls out of `wanted`, and it is invalidated like anything else. The refusal stands on three other legs instead:

- **It asserts completeness.** Every other derived fact says "these readings hold, therefore this". A sum says "these readings *are all the readings*, therefore this", and the second half is not a fact an open-world base can hold.
- **Its proof cannot explain its own retirement.** The withdrawn row still names A, B and C, and all three are still live. The reading that changed the answer was never one of its premises, so the proof is silent about it.
- **It forecloses incremental maintenance** (0002 R3, still open). An incremental pass walks premises; a new reading that is nobody's premise wakes nothing.

The index row is updated to match.

Issue #497, opened 2026-09-08
There is no way to zoom from a class to its instances


WaylandYang:
The ontology is the contract the graph obeys (0012). The interface shows the contract on one page and the data on another, and there is no way to get from one to the other except a single link buried in the class panel's Instances tab.

So a reader cannot ask the question the product is built to answer: *three classes and four relations produced how many facts?* The schema diagram says the ontology has `Person —works_at→ Organization`. The graph says `Lin Zhao —works_at→ Meridian Systems`. Nothing puts them in one frame, and nothing lets you zoom between them.

## What to build

A two-position level switch on the graph page, `Schema` and `Instances`, beside the search box, recorded in the URL as `?level=schema`.

**Shared across the levels**: the canvas (#496), the type legend — class colours already live on `entity_types` — the search box, and the docked panel's geometry.

**Instance level only**: the time scrubber, the inference popover and the derived-edge toggle. A class has no interval and is not derived, so those three retract at the schema level.

**The two crossings are the point:**

- Click a class at the schema level: switch to the instance level, filtered to that class. This is the drill-down from contract to data.
- Click an entity's type in the entity panel: switch to the schema level with that class selected. This asks which clause the fact is filed under.

**The schema level's panel is read-only**: definition, axioms, instance count, and a link to edit it in the ontology page. Reading happens on the graph page and authoring on the ontology page; the line does not blur.

## The diagram has one home

After this cut, `OntologySchemaGraph.tsx` is no longer a second page's canvas. It becomes the schema projection feeding the shared canvas, and the ontology page does not render a canvas at all (#ISSUE3). One picture, one home; the two pages reach each other by link, not by each drawing their own copy.

The rule goes in `web/DESIGN.md`.

## Not in this cut

Whether the schema level should show relation cardinality, instance counts on the edges, or a heat scale by usage. Land the level switch first and look at it.

WaylandYang:
Not doing this. Built in #508, looked at running, and cut: on the graph page `Instances` and `Schema` switch between two different datasets, so before you can read anything you have to decide which one you are looking at. The record's premise — that the contract and the data belong in one frame — survives; putting them behind a switch on the reader's page is not how to honour it.

The diagram's own defects that the cut turned up land in #511. The crossings it was for stay unbuilt: from a class to its instances, and from an entity to the class it obeys.

Issue #498, opened 2026-09-08
The ontology can only be read one class at a time


WaylandYang:
The ontology page reads one class at a time. A rail of names, a diagram, and a panel showing whichever one you clicked. Every question that spans the ontology is unanswerable there:

- which properties are functional, and which were never declared either way;
- which classes have no instances after a pack import brought in nine hundred;
- which relations have no domain, so the extractor gets no signature for them;
- which of them are events rather than states, now that the difference drives the engine (0031);
- what attributes this base has at all — attributes appear only inside a class's panel, one class at a time, so the base-wide list exists nowhere.

Every other page of this kind was rebuilt as a table this month: tokens, data sources, members, the deployment's users, and business rules (#482, which said in as many words that a column of cards did not look like anything else we had just rebuilt). The ontology is the last one standing, and it is the one with the most columns.

## What to build

Three tables behind a segmented control.

| | columns |
|---|---|
| **Classes** | name, with its colour dot and shape, indented by depth; key; parent; disjoint; instances; attributes; description |
| **Properties** | name; key; the subject and object signature; temporal semantics; the six axioms as chips; inverse; super-property; facts |
| **Attributes** | name; key; the class it hangs on; datatype; unit; single-valued; usage |

Attributes get their own table because the base-wide list has no home today.

The axioms are chips, not six checkboxes: six boxes in a row cannot be read as "what does this property guarantee", and a few letters can be scanned down a column.

A row opens the edit dialog that already exists (#480). The instance and fact counts are links into the graph page, filtered. Each row carries a "see it in the diagram" action that deep-links to the graph page's schema level (#497).

## The hierarchy in a table

`subClassOf` is the ontology's backbone and a flat list loses it. The rule: the default order is the hierarchy, with the name column indented, the way the rail's tree already indents. Sorting by any other column flattens the list, and the parent column is always present, so flattening loses nothing but the shape. That is what a file manager does, and it is honest about which of the two readings you are in.

## Two consequences

**The page loses its canvas.** The diagram lives on the graph page's schema level after #497. The ontology page is tables and dialogs.

**Usage counts come back.** The rail's tree deliberately omits them, and its comment gives the reason: counting and rendering are a burden once the numbers grow. `usage` is already returned on both views; the burden was rendering hundreds of rows at once, which paging solves.

## Needs one thing in `ui/`

[...]

PR #508, opened 2026-09-08
The schema is a level of the graph

WaylandYang:
Part of #497. Stacked on #499 — its shared canvas is what lets the schema diagram be mounted somewhere else.

The ontology is the contract the graph obeys (0012), and the interface showed the contract on one page and the data on another with one buried link between them. *Three classes and four relations produced how many facts?* had nowhere to be asked.

## What lands

A two-position switch beside the search box, `Instances` and `Schema`, recorded in the URL as `?level=schema` so a link carries it.

**Schema draws the ontology page's diagram.** Not a copy: `OntologySchemaGraph` was already a controlled component, so the graph page mounts the same one. It gained a single optional prop, `chromeTop`, because embedded it has to let the host page's floating bar go first.

**The chrome follows the level.** Gone at schema level: the time scrubber, the inference tower, the layout and camera towers, the type legend, the entity panel and the "showing N of M entities" read-out. A class has no interval, is not derived, and the schema diagram brings its own zoom tower and its own legend. The search box switches to searching classes; picking one brings it into view, the same path the ontology page's rail takes.

**Two crossings, which are the point.**

- *Contract to data.* Select a class, and the bar offers `Only Person 172`. It switches to the instance level with every other type hidden — through the existing type-visibility set, so the legend shows what was hidden and the existing "Show all" undoes it. No second filtering mechanism.
- *Data to contract.* The type on an entity panel is now a link. Clicking `Organization` on Sam Altman lands on `?level=schema&cls=organization` with that class selected and focused.

**Clicking a class selects it; it does not drill.** The first version drilled on click, and the schema level was then unusable for what it exists for — one click and you were in the other level. Selecting shows the class's relations with everything else muted, and drilling is the explicit second step.

The class key, not its id, goes in the URL: a `GraphNode` carries `type_key` and no class id, and a key reads better in a shared link.

## Three bugs this shook out, all worth the comments they got

- **The instance canvas came back blank.** Its effect depends on `data.data`; the schema level unmounts the container, and a ref appearing does not re-run an effect. `level` joins the dependency array.
- **The return crossing switched level and was immediately undone.** Clearing the selection fires the URL-sync effect, which wrote the whole search object back and dropped the `level` that `goLevel` had just set. It now merges instead of replacing.
- **The embedded legend and the host's search dropdown landed on each other.** The page's bar moves to `z-20`, above the canvas's own chrome.

## Verified in the browser

Against a copy of the benchmark base, 1415 entities and 916 classes: the switch both ways; the class search finding and focusing `Person`; `Only Person 172` landing on an instance canvas of exactly the people, with `Show all (915 hidden)` offered; clicking `Organization` on an entity panel arriving at `?level=schema&cls=organization` with the class selected; and the ontology page unchanged.

`pnpm build` clean: style guard, `tsc`, vite.

## Left for the next cut

The schema level has no panel. Selecting a class shows its shape on the canvas and offers the drill, but not its definition, axioms or description — those live on the ontology page, and whether a read-only version belongs here is worth looking at rather than guessing. The drill's semantics — "only this class", which leaves sparse edges because a class's neighbours are hidden with everything else — is the other thing to look at with real eyes.


WaylandYang:
Follow-up commit from looking at it in the browser. Four things, three of them things I had left open or broken.

**A selected class now says what it is.** The read-only panel the PR description called "left for the next cut" is here instead, because a schema level where selecting a class shows nothing is not worth looking at. `DockedPanel`, `PanelHeader`, `Def` and `Description` move out of `Ontology.tsx` into `pages/classPanel.tsx`, so the docked panel stays one shell rather than two — DESIGN.md rule 6 says there is only one kind of docked panel, and that stops being true the moment a second page draws its own. `SchemaClassPanel` is that shell with three read-only tabs: Definition (parent, disjoint, description), Relations (which relations this class takes part in, and with what on the other side), Instances (paged, and clicking one drops to the instance level on that entity — a third crossing, and the most useful one). Its single action is a link to the ontology page. Reading here, authoring there.

**The legend was on its own second row.** I had pushed the embedded diagram's chrome down with a `chromeTop` offset, which is how it ended up as a second line under the search box. It now goes through a portal into a slot the host page puts in its own bar, so search, level switch, drill button and legend are one row. The ontology page passes no slot and keeps its floating legend.

Two flexbox bites on the way: the slot with `min-w-0` shrank to zero width and the chips spilled down the canvas in a column; without it, `min-width: auto` on a `flex-wrap` child resolves to one chip wide and it still wrapped. `shrink-0` is what holds it on one line.

**The legend swatches read as uneven line weights.** They are all the same 2px, but they used the *focus* colours — inheritance at 95% white beside relations at 60% white. Two lines differing only in brightness read as "thick and thin", not "two kinds of edge". They are now four opaque colours at one weight, taken from each edge's resting colour, and the identity is carried by hue.

**The schema's edges were heavier than the instance level's.** `MIN_EDGE_THICKNESS` was 3. Beyond the mismatch across levels, that floor was flattening a distinction the diagram makes on purpose: inheritance edges are sized 0.9 and relation edges 2, and a floor of 3 drew both at 3. It is now 2, which matches the relation edge's own size, lets inheritance be thin again, and still leaves an edge you can hit with a mouse — the reason the floor exists.

`pnpm build` clean. All four looked at in the browser on the same base.

WaylandYang:
Closing unmerged. The level switch is out: on the graph page, `Instances` and `Schema` switch between two different datasets, and deciding which one you are looking at before you can read anything is a tax the page did not use to charge. Judged from looking at it running, not from the description.

What the cut found out along the way is worth keeping, and lands separately in #511 — every one of those was a defect in the schema diagram whether or not it is ever embedded anywhere:

- it drew every class that has an instance (142 here, most with a single instance), so `Person` at 172 looked no more important than an `ImageObject` at 1. It now draws the biggest classes, on the same four-step ladder the instance graph uses, with the stepper beside the `+N classes` pill that reports what is held back;
- `MIN_EDGE_THICKNESS` was 3, which flattened the diagram's own distinction — inheritance edges are sized 0.9 and relation edges 2, and the floor drew both at 3;
- the legend's swatches used the focus colours, so inheritance at 95% white beside relations at 60% white read as uneven line weights rather than two kinds of edge;
- the legend row mixed 4px chips and 6px pills at identical size;
- the force settings differed from the instance graph's for no recorded reason (gravity 0.55 against 0.35), so the same engine drew one graph open and the other balled up.

The read-only class panel and the `chromeSlot` portal go with the switch; `classPanel.tsx` keeps the docked shell, which the ontology page now imports rather than defining its own.

#498 is revised rather than dropped: the ontology page keeps the diagram and gains the tables as a second view of the same page, switchable, with its rail intact. The switch that is heavy on the graph page is not heavy there — table and diagram are two drawings of one thing, not two datasets.

PR #521, opened 2026-09-08
Wake idle workers when jobs are queued

### LINKED ISSUE #517: A queued job waits two seconds for a worker that is doing nothing
The worker loop polls, and when the queue is empty it sleeps two seconds:

```rust
// crates/utopia-store/src/jobs.rs:329
match claim_one(&pool).await {
    Ok(Some(job)) => { ...spawn... }
    Ok(None) => tokio::time::sleep(Duration::from_secs(2)).await,
    Err(e) => { ...; tokio::time::sleep(Duration::from_secs(5)).await; }
}
```

An idle deployment is the common case, so the common case for a newly enqueued job is that it waits. Upload a document and parsing starts up to two seconds later; that document then enqueues `extract_document`, which waits again. The user sees a document sitting in `pending`, then sitting in `queued`, for a delay that is pure polling latency and nothing else.

Shortening the sleep is the wrong trade — it buys latency with a constant background query against a pool that [`db.rs:5`](crates/utopia-store/src/db.rs#L5) sizes for bursts of short queries, and it never reaches zero.

Postgres already has the mechanism. `NOTIFY` on enqueue, `LISTEN` in the worker, and the loop waits on either a notification or a timeout instead of a bare sleep. Enqueue-to-start becomes immediate and the polling continues underneath as the backstop.

## Keep the poll

The timeout arm is not redundant and must stay:

- `jobs.run_at` schedules work in the future — the retry backoff at [`jobs.rs:255`](crates/utopia-store/src/jobs.rs#L255) is 30s, 120s, 270s. Nothing sends a notification when a job's `run_at` arrives, so only the poll finds it.
- A notification sent while the worker is between `LISTEN` calls, or during a connection drop, is lost. `LISTEN/NOTIFY` has no delivery guarantee across a reconnect, and sqlx's `PgListener` reconnects on its own.
- Multiple workers can be woken by one notification and only one will win `claim_one`. The losers must fall back to waiting, not spin.

So the change is "wake early when there is news", not "stop polling". With the poll retained the timeout could reasonably grow rather than shrink, which costs less than it does today.

## Where the notification goes

`jobs::enqueue` is the obvious site, but the notification has to be sent **in the same transaction as the insert**, or a notification can arrive before the row is visible and the woken worker finds nothing — the two-second poll then supplies the delay the change was meant to remove. `NOTIFY` inside a transaction is delivered on commit, which is exactly the required behaviour, so this is a matter of not moving it out.

## Not this

**Replacing the poll with `LISTEN` alone.** Delayed retries would never run and a dropped notification would strand a job until the next enqueue.

**An in-process channel instead of `NOTIFY`.** It works today because the deployment is single-process — the orphan reclamation at [`jobs.rs:283`](crates/utopia-store/src/jobs.rs#L283) says so plainly, requeueing everything `running` at startup on the grounds that anything still running must be an orphan. But that assumption is already the weakest thing in this file, and an in-process channel would weld it in. `NOTIFY` costs the same and does not.

## Unit tests

The SME rule is that the queue's guarantees are unchanged and only its latency moves. Every existing property — a job runs once, a failure backs off and eventually gives up, a delayed job does not run early — has to hold with a wake-up path added.

[`crates/utopia-store/tests/a_failed_job_finds_its_way_back.rs`](crates/utopia-store/tests/a_failed_job_finds_its_way_back.rs) is the existing guard for the retry path and must pass unchanged.

New:

- `a_queued_job_starts_without_waiting_for_the_poll` — enqueue against an idle worker and assert it starts well inside the poll interval. The point of the change, stated as a rule.
- `a_job_scheduled_for_later_is_not_woken_early` — a `run_at` in the future is not claimed by a notification. The backoff exists to space retries out and a wake-up must not collapse it.
- `a_delayed_job_still_runs_when_its_time_comes` — with no notification ever sent, the poll finds it. This is the arm that would be lost by going to `LISTEN` alone.
- `a_job_is_claimed_once_even_when_several_workers_wake` — two workers, one notification, one execution. `claim_one` already has to guarantee this; the change gives it a new way to be exercised.
- `a_lost_notification_costs_latency_and_not_the_job` — drop the listener connection, enqueue, and assert the job still runs on the poll.
- `a_notification_never_arrives_before_the_row` — enqueue inside a transaction that is held open, assert no worker sees it, then commit and assert it is picked up. This is the ordering hazard described above and the one worth pinning.
- `an_orphaned_job_is_still_reclaimed_at_startup` — the behaviour at `jobs.rs:283` is untouched.

The backoff arithmetic itself is already covered by a pure unit test at [`jobs.rs:341`](crates/utopia-store/src/jobs.rs#L341) and needs nothing new.

## Acceptance testing

1. Upload a document on an otherwise idle deployment and time upload → `parsing`, and `ready` → extraction starting. Both before and after, both in the PR. The two-second steps should disappear.
2. Watch the document status in the UI through a full ingest and confirm the transitions no longer stall between stages.
3. Leave a deployment idle for an hour with query logging on and confirm the background query rate has not gone up.
4. Force a job to fail and confirm the retries still land at roughly 30s, 120s and 270s and that it gives up after the third — the wake-up must not have collapsed the backoff.
5. Restart the server with jobs mid-flight and confirm they are reclaimed and rerun.
6. Kill the database connection under a running worker, bring it back, enqueue a job, and confirm it is eventually picked up.
7. `bash scripts/smoke.sh` against a throwaway deployment; `cargo test --workspace` with `UTOPIA_TEST_REQUIRE_DB=1`.

rksharma-owg:
Closes #517

## Summary

- add a PostgreSQL notification channel for queued jobs
- wait for either a job notification or the existing polling backstop when the worker is idle
- emit notifications transactionally for normal, deduplicated, extraction, RSS, and manual requeue paths
- add a database-backed regression test for the notification path

## Validation

- `cargo fmt --all --check`
- `UTOPIA_DATABASE_URL=postgres://utopia:utopia@localhost:1517/utopia UTOPIA_TEST_REQUIRE_DB=1 cargo test --workspace`
- `./node_modules/.bin/vite build`

The repository's `pnpm build` wrapper could not complete because this environment's pnpm policy requires explicit approval for the `esbuild` install script; the direct Vite build passed.


WaylandYang:
Merged, with two changes pushed to your branch first so it could land today:

- **The idle wait sleeps when the listener fails.** On a `recv` error the loop went straight back to `claim_one`; a database that accepts claims while the listener cannot reconnect (an exhausted pool is one way) would have turned that into a tight loop of one claim and one warning per iteration. The wait is now one function, `jobs::wait_for_work`, and an error sleeps the poll interval before the next claim.
- **The test exercises what the worker does.** The original asserted that a notification arrives. The new file drives `wait_for_work` itself, which is what the loop calls, and pins three rules: an enqueue ends the wait with `Notified` inside the poll interval; a row inserted in an open transaction wakes nothing until commit, which is the ordering hazard #517 names; with nothing queued the poll arm still fires. It stays on the wait function rather than `run_worker` because a live worker on a shared test database would claim other tests' jobs.

The notification sites are exactly as you wrote them, all five. One comment added: the listener keeps a pool connection for good, so the pool has one fewer to hand out. Thanks for the complete coverage; that is what decided between the two PRs for #517.

PR #522, opened 2026-09-08
Wake idle job workers when work is queued

### LINKED ISSUE #517: A queued job waits two seconds for a worker that is doing nothing

ousamabenyounes:
Fixes #517.

What changed:
- Send a Postgres notification in the same transaction when jobs are inserted or failed jobs are requeued.
- Have the worker LISTEN for that notification while keeping the existing 2 second polling fallback.
- Add DB-backed regression coverage for idle workers waking on both enqueue and requeue.

Verification:
- RED on clean `origin/dev` with the regression tests: both failed with `deadline has elapsed`.
- GREEN on this branch: `cargo test -p utopia-store --lib without_waiting_for_the_poll -- --nocapture`.
- GREEN focused DB checks: `cargo test -p utopia-store --lib`, `cargo test -p utopia-store --test a_judged_entity_waits_its_turn`, `cargo test -p utopia-store --test rss_full_content`, `cargo test -p utopia-store --test a_failed_job_finds_its_way_back`.
- Full local replay: `cargo test --workspace`, `cargo build --workspace`, and `cd web && pnpm install --frozen-lockfile && pnpm build` passed.

Note: local `cargo clippy --workspace --all-targets -- -D warnings` still fails on the same pre-existing Rust 1.93 `collapsible_else_if` warnings in `crates/utopia-store/src/temporal.rs:124` and `:142` that were present on the baseline before this change.

### DIFF
--- a/crates/utopia-store/src/jobs.rs
+++ b/crates/utopia-store/src/jobs.rs
@@ -5,12 +5,23 @@
 //! 目标数经 AtomicUsize 热读——系统设置里改并发即时生效，无需重启。
 
 use chrono::{DateTime, Utc};
+use sqlx::postgres::PgListener;
 use sqlx::{PgPool, Postgres, Transaction};
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
 use utopia_core::AppResult;
 use uuid::Uuid;
+
+const DEFAULT_MAX_ATTEMPTS: i32 = 3;
+const MAX_ATTEMPTS_LIMIT: i32 = 32;
+const MAX_ATTEMPTS_VALIDATION_ERROR: &str = "max_attempts must be between 1 and 32";
+const RETRY_DELAY_BASE_SECS: i64 = 30;
+const JOB_WAKE_CHANNEL: &str = "utopia_jobs_wake";
+const NOTIFY_JOB_WAKE_SQL: &str = "SELECT pg_notify($1, '')";
+const WORKER_FULL_SLEEP: Duration = Duration::from_millis(200);
+const WORKER_IDLE_POLL: Duration = Duration::from_secs(2);
+const WORKER_ERROR_RETRY: Duration = Duration::from_secs(5);
 
 #[derive(Debug, Clone, sqlx::FromRow)]
 pub struct Job {
@@ -29,6 +40,7 @@ pub async fn enqueue_unless_queued(
     kind: &str,
     payload: serde_json::Value,
 ) -> AppResult<Option<i64>> {
+    let mut tx = pool.begin().await?;
     let row: Option<(i64,)> = sqlx::query_as(
         "INSERT INTO jobs (kind, payload)
          SELECT $1, $2
@@ -37,13 +49,17 @@ pub async fn enqueue_unless_queued(
     )
     .bind(kind)
     .bind(payload)
-    .fetch_optional(pool)
+    .fetch_optional(&mut *tx)
     .await?;
+    if row.is_some() {
+        notify_job_wake_tx(&mut tx).await?;
+    }
+    tx.commit().await?;
     Ok(row.map(|(id,)| id))
 }
 
 pub async fn enqueue(pool: &PgPool, kind: &str, payload: serde_json::Value) -> AppResult<i64> {
-    enqueue_with_max_attempts(pool, kind, payload, 3).await
+    enqueue_with_max_attempts(pool, kind, payload, DEFAULT_MAX_ATTEMPTS).await
 }
 
 /// 入队并显式指定有限的重试预算。调用方不能把一个外部系统的任务变成无限重试。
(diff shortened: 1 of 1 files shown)


WaylandYang:
Closing in favour of #521, which landed with the same mechanism. Two things decided it, both about coverage rather than the idea:

- **Two enqueue sites are not in this PR.** `documents::queue_extraction_one` and `queue_extraction` insert into `jobs` directly, so `extract_document` would still wait out the poll, which is the example #517 itself gives (upload → parse → extract). `rss_full_content::requeue_failed` is the third. #521 notifies from all of them.
- **The tests reach outside their own rows.** `DELETE FROM jobs` clears the whole table, and a live `run_worker` with a no-op handler marks any queued job in the database done. Every database-backed test in this repository creates its own rows and removes only those; pointed at a developer's database these two would erase real work.

The worker-side shape of your tests was the better one, and #521 landed with a test of the same intent, a real wait that ends on the notification plus the commit-ordering rule, written to stay inside its own rows. Thank you for the PR.

PR #538, opened 2026-09-09
LLM errors preserve their response bodies

### LINKED ISSUE #527: An error body the client cannot parse is reported as unknown error
`err_detail` in `crates/utopia-llm/src/lib.rs:521` reads `error.message`, then `message`, and otherwise says `unknown error`. In #519 the endpoint's 400 carried the entire diagnosis:

```
request (105140 tokens) exceeds the available context size (32768 tokens)
```

and it reached the job log as

```
LLM request failed (400 Bad Request): unknown error
```

The reporter put a logging proxy between the app and the endpoint to read what we had already been handed. Whatever shape LM Studio used, it matched neither key. The proxy capture shows `error.message` present, so either the body arrived in another form or it was not the JSON we parsed; either way the fallback should quote what arrived, not discard it.

## The second way a body is lost

`chat()` at `:251` does `resp.json().await.map_err(Unreachable)?` before looking at the status. A non-JSON error body, which is what a proxy's HTML page or a plain-text 502 looks like, is reported as *unreachable*. That points at the wrong thing: the endpoint answered, and the answer was the diagnosis.

## Fix direction

Read the body once as text. Try JSON for `error.message`, `message`, and `error` as a plain string. If none applies, quote the raw text, trimmed and truncated (500 characters is enough to carry any diagnosis seen so far). Keep the status in front of it. `failure()`'s classification of `OutOfCredit` and `RateLimited` keeps reading the JSON where there is JSON; nothing about it changes.

## Tests

Pure unit tests beside the existing `failure` tests:

- `an_error_given_as_a_string_is_forwarded`
- `a_body_that_is_not_json_is_quoted_and_not_called_unreachable`
- `a_body_with_error_message_still_reads_it`: the existing behaviour, pinned
- `a_long_body_is_truncated`

Related: #519.

ousamabenyounes:
## Summary

- Check the HTTP status before decoding successful LLM responses as JSON.
- Preserve plain-text and unfamiliar JSON error bodies, capped at 500 Unicode characters.
- Keep quota and rate-limit classification for recognized JSON errors across chat, streaming, tool, and embedding requests.

Closes #527.

On unmodified `dev`, a local endpoint returning `502 Bad Gateway` with `proxy says the model context is too long` failed as:

```text
assertion failed: !is_unreachable(&error)
```

The same regression test passes after the patch and includes the endpoint diagnosis in the error.

## Validation

- `cargo test -p utopia-llm`: 17 passed, 0 failed.
- Fix removed temporarily: the regression test fails again with the original `Unreachable` classification.
- Changed executable production lines: 100% covered (28/28).
- Full local suite on upstream `dev` at `cb996aa`: migrations twice, formatting, workspace clippy with warnings denied, workspace tests, required DB-backed store tests, workspace build, and frozen-lockfile web tests/build all passed.
- Suite comparison: baseline 659 passed / 0 failed / 1 ignored; patched 691 passed / 0 failed / 1 ignored (the patched total includes the targeted coverage replay).

### DIFF
--- a/crates/utopia-llm/src/lib.rs
+++ b/crates/utopia-llm/src/lib.rs
@@ -62,15 +62,15 @@ pub enum ToolStreamItem {
 }
 
 /// **没能从端点拿到一个能解析的回答。** 两种：连不上（DNS、连接、TLS、超时），
-/// 或者连上了但回来的根本不是这个 API（响应体解不成 JSON）。
+/// 或者连上了但成功响应根本不是这个 API（响应体解不成 JSON）。
 ///
 /// 合成一类是有意的：从用户那边看这两种是同一件事——"你配的这个地址不是模型 API"，
 /// 该做的也是同一件事：去看 URL、看代理。第一版只收传输层失败，结果最常见的那种
 /// 故障（URL 配错、代理挡在中间回了 HTML）一条告警都不产生，
 /// 正是"失败无声"本身。
 ///
-/// **不含**端点干干净净地回了 4xx/5xx：那说明它就是模型 API，只是密钥、
-/// 配额或模型名不对——另一类问题，该找的人也不同。
+/// **不含**端点回的 4xx/5xx，即使错误体不是 JSON：那说明服务器已经回答，
+/// 错误体本身就是诊断——另一类问题，该找的人也不同。
 ///
 /// 有类型而不是匹配错误文本：调用链上任何一层加一句 context 都会改文本，
 /// 而 `anyhow` 的 source 链让 `downcast_ref` 一路都认得出来。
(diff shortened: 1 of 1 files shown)


WaylandYang:
Merged. This is the fix #527 asked for, in the shape it asked for: the body is read once as text, the status is checked before anything is parsed as JSON, the three JSON shapes are tried and the raw text is quoted otherwise, capped at 500 characters, and a body that cannot be read at all stays `Unreachable`. Updating the doc comment on `Unreachable` to say what it now excludes was the right call; that comment is what the next reader will trust.

The tests are the useful part: one raw-socket server per case, every request path covered for the non-JSON body, the incomplete body pinned as unreachable, and the multibyte truncation counted in characters. Thanks for reading the issue's test list and writing to it.

One habit to mention for next time, not a blocker: the constants pulled out of the tests and the production code (`LLM_REQUEST_KIND`, `TEST_USER_ROLE`, `SUCCESS_TEXT` and friends) add lines without adding meaning here; a literal at the point of use reads faster in a test. This repository leans toward the literal.

WaylandYang:
One thing you could not have seen from Linux: the raw-socket servers in the tests wrote their answer and closed without reading the request. Windows answers an unread socket with RST, so the client got "connection aborted" (os error 10053) instead of the body, and two of the six tests failed on a Windows checkout while passing on CI. #542 has them read the request first and shut down cleanly; behaviour under test is unchanged. Worth knowing for the next fake server. The same PR puts the literals back at their point of use, as mentioned above.

PR #541, opened 2026-09-09
Model probes report reachability and authentication

### LINKED ISSUE #528: Test connection passes for an endpoint that cannot run one extraction
`settings_routes::test` (`crates/utopia-server/src/api/settings_routes.rs:77`) sends the chat model `Reply with exactly one word: OK` and the embedding model the string `connectivity test`. It certifies that the endpoint is reachable, the credentials work and the model name resolves. It says nothing about whether the model can run one extraction, and the button's name suggests it does.

In #519 the probe passed against a 32k endpoint, and every extraction chunk then failed with a 400 on prompt size. The reporter's observation: "`Test connection` passing is misleading. It exercises a trivial completion, so it certifies an endpoint that cannot run a single extraction."

## What it should say

The smallest change is honest wording: the result reports what was tested, so a passing probe reads as *reachable and authenticated*, not *ready*.

Beyond that, when the endpoint reports a context length, compare it with what a chunk actually costs. With per-chunk retrieval a chunk carries roughly 100 to 150 ontology terms plus instructions, 6 to 8k tokens by the constant measured in #519 (~48 tokens a term). An endpoint under that gets a warning naming the number. Sending a probe of that size is not the answer; on a hosted model it is a real bill for every click.

## Priority

Low. Once extraction fails with a reason of its own (#519's fix and the follow-ups), the probe stops hiding a failure and merely stops short of predicting one. Do it after those.

Related: #519, #470 (per-task models, which would let extraction point at an endpoint chosen for this).

ousamabenyounes:
## Summary

- Describe successful chat and embedding probes as establishing reachability and authentication.
- Keep the returned reply and embedding dimension so the result still shows what the probe observed.
- Apply the same honest wording in English and Chinese.

Closes #528.

On unmodified `dev`, the new copy contract failed because the result was only `Connected (OK)`:

```text
Expected: "Reachable and authenticated (OK)"
Received: "Connected (OK)"
```

After the patch, all 9 web tests pass. Temporarily restoring the old formatter reproduces the same RED failure.

## Validation

- Changed executable production lines: 100% directly invoked and asserted (4/4).
- Full local suite: migrations twice, formatting, workspace clippy with warnings denied, 668 Rust tests passed / 0 failed / 1 ignored, required DB-backed store tests, workspace build, 9 web tests, and the web production build all passed.
- Two independent Codex reviews, including the resolved OCR rule set, accepted the final diff.


WaylandYang:
Merged. This is the wording #528 asked for: a passing probe now says what it tested, in both languages.

Two things changed on your branch before the merge, pushed as a commit of their own so the history shows who did what. The Chinese result now reads 已连通，认证通过, the same statement in two clauses. And `settingsProbeCopy.test.ts` is gone: a test that asserts an i18n string equals itself covers no behaviour and doubles the places a copy change has to touch, so this repository keeps copy free of tests. Thanks for the fix.

Issue #526, opened 2026-09-09
Extraction holds a worker while the ontology index is built


WaylandYang:
#523 makes `extract_document` call `ontology_index::refresh` before per-chunk retrieval whenever the ontology is over `ontology_prompt_budget`. That closes the fallback #519 hit, where retrieval found no vectors and the full vocabulary went out at 105k tokens a chunk. The wait it introduces is the right wait. Where it happens is not: inside a running job.

## Three costs of waiting inside the job

**It holds a worker.** `refresh` is serialised per base (`ontology_index::PER_KB`), so every extraction that arrives while the index is being built parks on that lock with its worker slot. `worker_concurrency` defaults to 32 and the column's CHECK caps it at 32 (`0001_core.sql:134`). Upload a batch of a hundred documents into a fresh base with schema.org and the first 32 extractions occupy every slot for as long as the embedding takes; the other 68 parses, and every other base's jobs, wait behind them. On a local embedding endpoint, which is the deployment #519 describes, that is minutes.

**The document says `extracting` while nothing is extracted.** `set_graph_status(.., "extracting")` runs at `extraction.rs:595`; the refresh runs at 680. A user watching the Library sees 32 documents extracting and no facts arriving, with nothing that names the ontology as the reason.

**A partial index is used as if complete.** `refresh_scoped` returns `Ok(done)` when some batches fail; only a warning says so. #523 then extracts against whatever fraction got embedded, and the candidates per chunk come only from that fraction. The argument #523 makes for waiting at all, that a document extracted while the ontology was invisible is never re-extracted, applies just as much to half-visible.

## The pipeline already has the right shape

`pipeline.rs` queues `extract_document` only after the chunk vectors are written (`:88`, `:151`). The ontology's vectors are the same kind of prerequisite, and the same rule fits: extraction is not runnable until the vectors it retrieves against exist. So the wait belongs in the queue, not in the worker.

## Fix direction

The jobs table has no way to say "not yet". `claim_one` spends an attempt on claim (`jobs.rs:156`), and `mark_failed` backs off 30s × attempts² with a budget of three, so an extraction that returns `Err` while waiting is `failed` after about seven minutes. One primitive is missing:

[...]

Issue #556, opened 2026-09-09
Metric and Dimension retire, and a mapping is rendered from an alignment and a rule


WaylandYang:
Once a table is aligned and a definition is a rule, `Metric` and `Dimension` have no work left, and `concept_mappings` has a new source. [0036](../blob/dev/docs/decisions/0036-exploration-aligns-a-schema-to-the-ontology.md) decisions 4 and 5.

## The two classes retire

`ensure_concept_types` goes. They were admitted by #231 to stop exploration from silently producing nothing; the record's revision note on 0009 says why that fixed the symptom and kept the mistake. No exploration inserts a class without a proposal.

Existing bases have entities of these classes — twelve real concepts and twenty-eight column names on the wide bench base. The column names are #553's leftovers and go with it. The concepts become what they are under the record: an attribute of a real class, or a rule.

## A definition is a rule, and a shared convention is one rule others read

GMV is `sum(Order.paid_amount) where Order.is_valid`. `Order.is_valid ← status ∈ {paid, shipped, done} ∧ ¬is_test` is written once and read by GMV, net sales, order count, refund rate (0030). This is where "test orders don't count" lives — a condition on an attribute, with a name, changed in one place. On the wide corpus the absence of that one sentence moved chat from 17 of 18 right answers to 1 of 18 (#520).

The unit lives on the attribute (`relation_types.unit`), the conversion on the alignment (#555). 0032 noted nothing checks units when an expression is written; this gives units a place to come from.

## `concept_mappings` is rendered, not dropped

Its status, revisions, audit stream and page (0011 §2–3) are the review flow this still needs, and the benches (#501, #520) read it and score it as they do today. What changes is where a row comes from: rendered from a rule, the table it aligns to (#554), the column expressions and the source's dialect, into the `sql` column it has now. A rendered row names the rule it came from; a hand-edited one does not, and the two stay distinguishable. #552 (bulk confirm, external) touches the page only and is unaffected.

`derived` (0011 §1) was for "conversion rate = orders / visits"; under the record that is a rule reading two rules and the flag has no meaning. Kept until rendering exists, dropped with the consolidating migration.

## Not in this issue

Where a rule over aligned attributes **executes**. 0032 refused aggregation on the graph because the graph is open-world; a database table is closed-world, so `sum(Order.paid_amount)` is honest there — but that is another executor. Rendering a row for chat to use does not need it. Building it is the record's largest open question and gets its own issue when it is time.

Issue #554, opened 2026-09-09
Exploration proposes an alignment per table


WaylandYang:
Exploration today reads a schema and produces entities of two classes it inserted itself. [0036](../blob/dev/docs/decisions/0036-exploration-aligns-a-schema-to-the-ontology.md) decisions 1, 2 and 6: it produces instead an **alignment proposal per table** — the class the table is a table of, the attributes its columns carry, the relations its foreign keys are, and for each aligned column an expression that turns the column into the attribute's value — and a person adopts it.

## What a proposal holds

For `dw.dwd_ord_dtl` on the wide corpus:

| column | is |
|---|---|
| `amt_pay`, `qty`, `ord_st`, `dt_crt` | attributes of `Order` |
| `buyer_id`, `shop_id` | relations `Order → Customer`, `Order → Shop` |
| `shop_nm`, `prov`, `buyer_lvl` | attributes of `Shop`, `Address`, `Customer`, flattened in |
| `ver`, `etl_dt`, `rmk`, `price_old` | nothing |

Exploration finds existing classes and attributes first (the ontology, with a pack from 0008, usually has `Order`) and proposes new ones only where none fits — 0003's loop with a schema in place of a corpus. On a base with no ontology it proposes the class (decision 6); the difference from today is that a proposal waits and `ensure_concept_types` did not.

## Where it goes

`ontology_proposals` already has the four sections this needs — `entity_types`, `attribute_types`, `relation_types`, `map_to` — and the attribute payload already carries `datatype` and `unit`. What it does not have is a way to say "these forty rows are one table's alignment, adopt or reject them together", and `decide_proposal` today only marks a row; creating the ontology rows is the separate adopt endpoints' job. So this needs:

- an alignment table: table → class, column → attribute (or relation) with the conversion tree from #555;
- a proposal that groups a table's rows, adopted as one thing — adopting `Order` while rejecting every attribute of it is not a decision anyone means to make;
- an adopt path that writes the ontology rows and the alignment rows in one transaction.

## What exploration does not propose

Definitions. GMV, refund rate, valid orders — 0 of 18 on the wide corpus, and what it would propose is `sum(amt_pay)`: runs, looks right, off by a hundred and includes the test orders. A definition is a rule a person writes over aligned attributes (decision 4, #556). #507 is the other way a rule can arrive — stated in a document, with provenance — and is complementary, not this.

## Measured by

[...]

Issue #520, opened 2026-09-08
A chat answer is never checked against the data


WaylandYang:
#501 measures the proposals the explorer makes. Nothing measures the thing those proposals exist for: whether a person asking a question in chat gets the right number back.

The two are not the same measurement and one does not imply the other. A confirmed definition can be perfect and the answer still wrong — the model can pick the wrong source, join badly, filter on the wrong date column, or ignore the semantic layer and guess from the schema document. Conversely a base with no confirmed mappings at all still answers questions, by reading the schema document and writing its own SQL, and sometimes it is right. **Today there is no number for either case.**

## What is checkable

`query_data` records very little: its step is `{kind, label: source, detail: purpose}` — the SQL it ran and the rows it got go only into `tool_exchange`, truncated, as the message replayed to the model. So the material for scoring a turn is:

- the assistant's final text,
- the tool calls in `tool_exchange`, which carry the **SQL the model actually ran**.

That second one is the useful one. **Re-run the SQL the model ran and compare its result with the gold query's** — the same numbers-not-names judgement #501 uses, and it survives the truncation because the SQL is short even when its output was cut.

Then two columns rather than one:

- `sql_right` — the SQL it ran computes what the question asked
- `answer_right` — the number it printed is that number

They come apart. A model that runs the right query and then rounds, mislabels a unit, or narrates a different figure is wrong in a way a SQL-only score would call right, and that is exactly the failure a person reading the answer would suffer.

## Corpus

Reuse `truth/tpch.mappings.json`. Each of its 24 definitions is already a question: "what is the total discounted revenue", "how many orders are there", "what is the stock value at cost". The gold SQL is already written. That gives 24 single-value questions for free, in `truth/tpch.questions.json` keyed by the same `id`, so a failed answer points at the definition it needed.

A second tier — TPC-H's own 22 queries, which group and order and return many rows — comes after. Their business questions are in the specification; scoring compares result sets, sorted. Single-value first because it is the case where "the answer is a number" is unambiguous, and because it is the tier that closes the loop below.

## The loop this is for

The point is not one score. It is:

1. run the questions, see which are wrong;
2. for each wrong one, ask whether the definition it needed has a **confirmed** mapping — the script can answer this, since questions and mappings share an id;
3. if not, that is a coverage gap: explore again, or write the mapping by hand;
4. if it does have one and the answer is still wrong, the fault is in the prompt, the tool, or the definition itself;
5. repeat.

[...]

PR #561, opened 2026-09-09
The schema document is searched, not extracted

### LINKED ISSUE #553: The schema document is extracted as if it were prose
Mounting a data source ingests its schema as a markdown document under the per-base "Data schemas" folder, so chat can `search_chunks` for a table. That document then goes through extraction like any other, and the extractor — shown the `Metric` / `Dimension` classes exploration had just created — files every column name as an entity.

Measured on the wide corpus of #501: forty concept entities in the base, twelve from exploration, **twenty-eight column names** (`amt_pay` as a Metric, `buyer_id` as a Dimension, `dw.dim_shop` as a Dimension with eight facts on it). Timestamps line up: 23:59:36 is the mount, 00:01:26 is exploration.

[0035](../blob/dev/docs/decisions/0035-exploration-aligns-a-schema-to-the-ontology.md) decision 7: the schema document is a search corpus, not a source of facts. It keeps its place in retrieval and leaves extraction. This is the one decision in that record that is a bug fix and can land on its own.

## Shape

A document that is retrieved from but not learned from is a property of the document, set by whoever ingests it. `sync_schema_doc` is the only such ingester today. `pipeline::process_document` queues `extract_document` after the chunk vectors are written; a document marked this way stops there, with a `graph_status` that says so rather than `none`, so the Library does not show it as waiting.

Not a source-kind rule ("folders named Data schemas never extract"): that is a name standing in for a type guarantee, the defect class 0009 warned about.

Touches the ingest path that #526 also wants to move (the queue waiting for ontology vectors); the two are separable — this one decides *whether* a document is extracted, #526 decides *when*.

WaylandYang:
Closes #553. Implements decision 7 of 0035 (the record itself arrives with the bench branch; its status line gets this PR's number there).

Mounting a data source ingests the schema as a markdown document so chat can `search_chunks` for a table. That document then went through extraction like prose, and the extractor filed every column name as an entity — on the wide bench corpus, 28 of 40 concept entities were column names.

## What changes

- **`sources.config.extract = false`** marks a source whose documents are searched and never extracted. `Source::extracts()` reads it (default true; a non-boolean counts as unset, so a typo cannot silently stop a source). `sync_schema_doc` creates the "Data schemas" folder with it; migration `0045` backfills existing folders and adds `skipped` to `graph_status`.
- **`pipeline::process_document`** stops after indexing and embedding for such a document and marks it `skipped` — `none` would read as "not yet queued" in the Library, and it never will be.
- **`documents::queue_extraction`** filters these documents in SQL, so full rebuild, per-source re-extract and any future caller all skip them in one place. The two manual entry points (`POST /documents/{id}/extract`, `POST /kbs/{id}/sources/{sid}/re-extract`) refuse with `source_not_extracted` so a person hears why rather than "queued 0".
- Library: a `Not extracted` / `不抽取` label for the status; the per-source re-extract button is hidden when the source does not extract (the per-document button's status whitelist already excluded it).

Not a name rule ("folders called Data schemas never extract"): the flag is on the source, which is the thing that has the property.

## Verified

- Unit: `Source::extracts` over `{}`, `{path}`, `true`, `"no"`, `0`, `false`.
- DB test `a_schema_document_is_searched_not_extracted`: full-base and per-source `queue_extraction` leave the schema document out.
- End to end on the bench base: fresh KB, mount `wide` → `wide-schema.md` is `ready / skipped`, 0 entities, 0 extract jobs; both manual endpoints return 422 `source_not_extracted` and queue nothing.
- Migration applied on the bench base: both existing "Data schemas" folders backfilled, CHECK widened.

Touches the ingest path #526 also wants to change; separable — this decides *whether* a document is extracted, #526 decides *when*. Does not touch `Mappings.tsx` (#552) or `chat.rs` (#548).

### DIFF
--- a/crates/utopia-core/src/models.rs
+++ b/crates/utopia-core/src/models.rs
@@ -142,6 +142,61 @@ pub struct Source {
     pub created_at: DateTime<Utc>,
 }
 
+impl Source {
+    /// 这个来源下的文档要不要进抽取。
+    ///
+    /// **缺省是要。** 只有 config 里 `{"extract": false}` 明说了才不抽——schema
+    /// 文档就是这样（0035 决定 7）：它是给问数检索表结构的语料，不是事实的来源，
+    /// 进抽取的结果是每个列名变成一个实体。开关记在来源上而不是文档上，因为
+    /// 「只检索、不学习」是这一整个来源的性质；也没有拿来源的名字当规则，那是
+    /// 命名约定冒充类型保证（0009）。
+    ///
+    /// 值不是布尔的按没写处理：一个手滑不该让一整个来源静默停抽。
+    /// `documents::queue_extraction` 里的 SQL 判的是同一件事，改一处要改两处。
+    pub fn extracts(&self) -> bool {
+        self.config
+            .get("extract")
+            .and_then(serde_json::Value::as_bool)
+            .unwrap_or(true)
+    }
+}
+
+#[cfg(test)]
+mod source_extracts_tests {
+    use super::Source;
+
+    fn with(config: serde_json::Value) -> Source {
+        Source {
+            id: uuid::Uuid::nil(),
+            kb_id: uuid::Uuid::nil(),
+            kind: "folder".into(),
+            name: "x".into(),
+            config,
+            icon: None,
+            sync_interval_minutes: None,
+            sync_cron: None,
+            last_sync_at: None,
+            last_sync_status: "never".into(),
+            last_sync_error: None,
+            last_sync_added: 0,
+            ingest_token: None,
+            created_at: chrono::Utc::now(),
+        }
+    }
+
+    #[test]
+    fn only_an_explicit_false_turns_extraction_off() {
+        // 老来源的 config 是 `{}`，watch_folder 的是 `{"path": …}`：都照旧抽取
+        assert!(with(serde_json::json!({})).extracts());
+        assert!(with(serde_json::json!({ "path": "/x" })).extracts());
+        assert!(with(serde_json::json!({ "extract": true })).extracts());
+        // 不是布尔的按没写处理，而不是按 false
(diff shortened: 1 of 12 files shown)


WaylandYang:
Bench after this change, wide corpus, same scripts and truth as #501 / #520, fresh base each time.

| | before (#501 run, base 01a08375) | after #553 (base 01a08642) |
|---|---|---|
| concept entities on the graph | 93 (mostly column names, accumulated over runs) | **12**, all from exploration |
| schema document | extracted | `skipped` |
| exploration: metrics right / wrong | 0 / 5 | 0 / 6 |
| exploration: definitions covered | 0/18 | 0/18 |
| chat, no confirmed definitions | 1/18 (6%) | 2/18 (11%) |

The graph number is the one this PR is for, and it moved from 93 to 12. The accuracy numbers did not move, and were not expected to: chat writes SQL from the schema *document*, which is unchanged, and exploration still cannot see a convention that is not in the schema. The one extra right answer (`test_orders`) is inside single-run noise. The lever for those two rows is #554–#556; the seeded upper bound on the same corpus is 17/18 (#520).

WaylandYang:
Review before merge found one gap: the migration marked the source but left the documents alone, so on an existing base a schema document that had failed extraction stayed `failed`, and "Retry failed" (`retry_failed` → `queue_extraction_one`, which has no source filter) would send it back through extraction. The migration now backfills `none` / `failed` documents of a non-extracting source to `skipped`. `done` ones are left as they are: their column-name entities go away with a rebuild, which already skips the source.

Also noted, not changed: `memory_ingest` queues extraction without asking the source, so `{"extract": false}` set on the memory source by hand would not hold there. The flag is meant for mounted schema folders; wiring it into the memory path can wait for a reason to.

Issue #555, opened 2026-09-09
A column becomes an attribute through the tree a rule computes with


WaylandYang:
An aligned column is rarely the attribute's value as stored. `amt_pay` is in cents; `chnl` is a code (`1` means the app); `dt_crt` is an instant and the attribute wants a month. The conversion has to be written somewhere, and [0036](../blob/dev/docs/decisions/0036-exploration-aligns-a-schema-to-the-ontology.md) decision 3 says where: **the expression tree 0032 built for rules**, `{attr} | {const} | {op, l, r}` in `utopia-reason`, extended with the node kinds alignment needs.

## Why not SQL text

Today's `concept_mappings.expr` is SQL text, and it fails on two counts 0032 already made about rule expressions:

- it binds to a dialect — `FILTER (WHERE …)` is Postgres, and there are five engines (0018);
- it can name a column that does not exist, so the error moves from unfillable to undiscovered.

A tree is rendered per dialect at query time, and its leaves are attribute ids, so a reference to nothing fails at save time where a person is looking. Natural language is not an option either: not executable, therefore not verifiable — #520's baseline is what happens when the model re-translates a sentence every time.

## What the tree gains

Driven by corpora, not designed ahead. The wide corpus needs three node kinds rules do not have: a cast (`amt_pay / 100` is integer arithmetic today), a `case` over a code column, a date truncation. The next corpus says what else. `MAX_EXPR_DEPTH` stays.

## Text in, tree stored

0032 allowed "a box that parses into this same tree", and this is the case for it: a person and a model both write `amt_pay / 100` more readily than they compose it from pickers. `sqlparser` is already a dependency (the read-only gate in `query_engine`); it parses the text into the tree, and what it cannot parse into a supported node is refused. What 0032 forbade — storing the string and evaluating it at run time — stays forbidden.

## Open in the record, to be settled here

Whether `CASE chnl WHEN 1 THEN 'app' … END` is a conversion (this tree, on the alignment) or a dimension (a rule that names groups). The wide corpus has both readings of `chnl`, and exploration proposed exactly that expression as a dimension in #501's third run.

Used by #554 (the alignment stores this tree) and #556 (rendering walks it).

Issue #503, opened 2026-09-08
Twelve proposals and eighty columns look the same


WaylandYang:
`explore_mappings` cuts twice and reports neither.

The schema text stops at `MAX_SCHEMA_CHARS` (12 000) — and the `break` there leaves only the current source's column loop, so with several sources each one overruns separately and the constant is not the bound it reads as. Proposals stop at `take(12)`, a fixed number whether the source has three tables or three hundred.

Nothing downstream says how much was covered. The only signal is the alert raised when `accepted == 0`. Eleven proposals against an eighty-column wide table and eleven against a complete small schema render identically on the page.

## What to add

- A bound that scales with the source — `min(tables × 3, 60)` or similar — instead of a fixed twelve, and a cap on the schema text that actually holds across sources.
- Record the run: tables scanned, columns scanned, proposals accepted, which tables got at least one. Show it on the Data Mappings page.

This is the only signal that would reveal a gap once confirmation stops passing through a person, and it is what #501 needs to report `covered` against a real base rather than only against the bench corpora.

WaylandYang:
Landed in dev with #567:

- the cap scales with the source: `proposal_cap(tables) = (tables × 3).clamp(12, 60)`, and the schema text bound holds across sources;
- every exploration writes a row to `mapping_exploration_runs`: sources, tables and columns scanned, whether the schema was truncated, the cap, proposals returned and accepted, what was dropped and why (`source`, `kind`, `type`, `definition`, and since review of that PR also `cap` for what the bound cut and `decided` for proposals that landed on a confirmed or rejected definition), and which tables got at least one definition; a failed run is a row with an error, not an absence;
- the mappings list (`GET /kbs/{id}/mappings`) carries `last_run`, which is what the bench reads to score coverage.

Still to do here: show the run on the Data Mappings page (the number and the table list). That is the interface cut, after #552 settles the page.

Issue #501, opened 2026-09-08
A mapping proposal has never been scored


WaylandYang:
`explore_mappings` reads a source's schema, asks the model for up to twelve concepts and writes them as proposals. Nobody knows how good they are. There is no corpus, no truth, no score — while `auto_type_resolution` defaults to on because it was measured (39/41) and governance defaults to off because it was not. Every argument about confirming mappings automatically is currently an argument about a number nobody has.

## Scoring an open generation

`govern.mjs` scores a binary verdict keyed by two names. A mapping proposal has no such key: the concept name is the model's own invention, so "Paid revenue" matches no truth entry written as "已支付 GMV".

**Align on numbers instead.** A truth entry is a business definition plus one gold SQL. Scoring runs the proposal — `SELECT <expr> FROM <table>`, or `sql` verbatim — against the same source and compares the result with the gold query's. Same number, same definition, whatever it is called. No per-proposal hand labelling, and the judgement is objective rather than a matter of reading two names.

Details that need deciding: float tolerance; a proposal matching several truth rows counts once and the rest are duplicates; dimensions carry no number, so they are judged by whether the column is in the truth's group-by set and whether the explorer filed it as a metric.

## Four columns

- `covered` — of N truth definitions, how many some proposal computes
- `right` — of K proposals, how many run and match a truth definition
- `wrong` — **run and match nothing**, reported with the number produced and the nearest truth number
- `broken` — do not run at all (no such column, syntax)

`wrong` is the number that decides whether automatic confirmation can ever be default-on. A broken proposal is harmless: it fails loudly and a person sees it. A proposal that runs and is wrong carries the whole risk, because chat prints a number that looks entirely normal.

## Corpora: TPC-H first, then TPC-DS

Both are TPC's decision-support benchmarks, and what makes them useful here is not their data generator but their **published queries**: TPC-H's 22 and TPC-DS's 99 each state a business question and answer it in SQL. `sum(l_extendedprice * (1 - l_discount))` is not somebody's guess at what "revenue" means in this schema — it is the benchmark's own definition. That is exactly the shape a truth entry needs, and nobody can suspect the truth of having been written to match our implementation.

- **TPC-H** — 8 tables, a clean star with real foreign keys (`l_orderkey → o_orderkey`, `o_custkey → c_custkey`, `c_nationkey → n_nationkey`). Measures the basics and key recognition, and the customer → nation → region chain is the smallest case of one object whose fields live in three tables.
[...]

PR #567, opened 2026-09-09
The semantic layer gets measured, and exploration learns to align

WaylandYang:
Three measuring benches for the semantic layer, the coverage cut of #503, and decision record 0036.

## Benches (`scripts/bench/`)

- **`mappings.mjs`** (#501) scores what exploration proposes. Judgement is by **number**: a truth entry is a definition plus one gold SQL, the proposal is run against the same rows, same number means same definition whatever the model named it. Four columns — `covered`, `right`, `wrong`, `broken` — and `wrong` (runs, matches nothing) is the one that decides whether automatic confirmation could ever be default-on. Two corpora: **TPC-H** (schema and definitions from the specification's 22 queries, rows from `generate_series`, comments as a separate file so their value is measurable) and **`wide`**, a flattened e-commerce order table whose every definition carries a convention the columns do not state.
- **`ask.mjs`** (#520) scores chat end to end: it asks each question, re-runs the SQL the model actually ran, and compares with the gold — two columns, `sql_right` and `answer_right`, because they come apart. `--replay` re-scores from the conversations already in the base (seconds, not forty minutes); `--seed` writes the truth as confirmed definitions (the upper bound); `--confirm` confirms what exploration proposed (the product path); `--parallel N` asks several at once.
- **`lib.mjs`** holds the one `same()` both benches use, so their numbers are on one scale.

Numbers, recorded on #501 and #520:

| | exploration covered | chat right |
|---|---|---|
| TPC-H, none | 8/24 after #503 (6/24 before) | 23/24 |
| TPC-H, exploration confirmed | — | 24/24 |
| wide, none | 0/18 | 2/18 |
| wide, conventions as one page of prose | — | 14/18 |
| wide, truth seeded | — | 17/18 |

## #503 — an exploration says what it covered

`proposal_cap` scales with the schema instead of a fixed 12 (TPC-H coverage 25% → 33%); the schema-text cap is now a real cross-source bound; a run writes a row in `mapping_exploration_runs` (tables and columns scanned, truncation, cap, returned, accepted, drops with an example each, tables covered, error); `last_run` on `GET /kbs/{id}/mappings`. Single-source bases accept whatever `source` the model echoes — the first bench run lost all twelve proposals to a source name with a timestamp in it. Migration `0044`.

## 0036 — Exploration aligns a schema to the ontology

Exploration today inserts two classes and makes concepts entities of them; the wide base had 93 concept entities, most of them column names. The record moves the concept: an attribute of a real class, or a rule over such attributes; exploration proposes an **alignment** per table through `ontology_proposals`; conversions are 0032's expression tree; `concept_mappings` becomes rendered. Revision notes on 0009 and 0011. Decision 7 is implemented in #561; the written-definition door is #563; the conventions-as-prose dead end was measured at 14/18 and revised in place. Follow-ups: #554, #555, #556.

Also renumbers the exploration-ledger migration to `0044` (dev took `0041`) and merges dev.


WaylandYang:
Review before merge. Two ledger gaps fixed, one bench guard, one renumber:

- **The cap was invisible.** Proposals past `cap` were sliced off before the loop with no drop reason, so `returned − accepted ≠ Σ dropped` whenever the model returned more than the cap, which is exactly the wide-table case the ledger exists for. The excess is now recorded under `cap` with the count and the cap value.
- **A decided definition counted as accepted.** `propose` upserts only `WHERE status = 'proposed'`, but the loop counted every call. A second exploration on a base with confirmed or rejected definitions reported `accepted = N` while Review gained nothing. `propose` now returns whether it wrote (`rows_affected`), and a proposal that lands on a decision is recorded under `decided`. Coverage still counts the table: it has a definition, which is what coverage asks. Test `a_rejected_mapping_does_not_come_back` asserts the flag.
- `finish` failing is logged instead of swallowed.
- `lib.mjs`: `BENCH_PSQL` without `-d` used to make every `onDb()` fall through to the command's default database, `DROP SCHEMA` included; `-d` is now inserted when absent.
- The decision record is 0036: dev took 0035 for the vector index (#566) while this was open. README rows read 0032 … 0036 in order once this merges.

Not changed: migrations here use plain `CREATE TABLE`, like 0038–0042 on dev; CI's second `sqlx migrate run` skips applied versions, so `IF NOT EXISTS` only matters for a renumber after a base has run the old number, which 0044 has not.

PR #571, opened 2026-09-09
Exploration describes the data and asks what it cannot tell

### LINKED ISSUE #570: Exploration describes the data and asks what it cannot tell
Chat without conventions answers 2 of 18 questions on the wide corpus; with the six conventions written as one page of prose it answers 14 (#520). The conventions have to come from somewhere, and today there is nowhere to put them: not a class, not a mapping row, not a column of `knowledge_bases`.

Two halves, two provenances:

- **What the schema states** — what each table is a table of and what one row is, which columns are keys, the unit of a money column where a comment says so, what a status code means where a comment says so, the time axis, which two columns look alike and how they differ. Exploration reads the schema anyway; it can write this, and rewrite it each run.
- **What the schema does not state** — test rows excluded, which statuses count, which of two amount columns is the figure, whether net subtracts refunds. Exploration cannot write this (0 of 18 definitions on the same corpus), **and must not guess it**: a guessed convention in the prompt is followed as confidently as a right one, which is the automatic-confirmation failure in a new place.

So exploration writes a description and a list of **questions** — the conventions it needed and could not find, phrased so the owner answers each in a line. The model already knows what it does not know: asked for valid orders with no definition, it said in its own words that it was searching for one. A blank page nobody fills in becomes four questions anybody can answer.

## Shape

Three columns on `knowledge_bases` (migration `0047`), two writers:

- `data_description`, `data_questions` — written by exploration, every run.
- `data_conventions` — written by a person, through `PATCH /kbs/{id}`; exploration never touches it. Mixing them in one field would let the next run overwrite the answers.

Chat reads the description and the conventions into the system prompt, every turn, rather than relying on retrieval finding a document (the prose experiment hit on every question because every question was about a metric). Exploration reads the conventions into its own prompt, so a proposal can carry the filter and the unit conversion — which is the first thing that could move its 0 of 18.

## Measured by

`ask.mjs` / `mappings.mjs --conventions` write the truth file’s `conventions` array through the PATCH. Three numbers: chat with the generated description alone; chat with description and conventions (against the prose 14/18); exploration coverage with conventions in its prompt (against 0/18).

## Interface, after

A `SettingsCard` on the base settings: the description read-only with Regenerate, the conventions editable, the questions listed underneath so answering them is the way to write conventions. Follows `web/DESIGN.md`.

WaylandYang:
Closes #570. Stacked on #567 (it changes `explore_mappings`, which #503 already rewrote there); retarget to `dev` once #567 lands.

Chat without conventions answers 2 of 18 questions on the wide corpus; with six conventions written as prose it answers 14 (#520). The conventions had nowhere to live. This gives them a home and lets exploration write the half it can.

## What changes

- `knowledge_bases` gains three columns (migration `0047`): **`data_description`** and **`data_questions`**, written by exploration every run; **`data_conventions`**, written by a person through `PATCH /kbs/{id}` and never touched by exploration. Two writers, two paths (`kbs::set_data_description` vs `kbs::update`); mixing them in one field would let the next run overwrite the answers.
- Exploration makes one more model call before proposing: a description that states only what the schema and comments say (grain, keys, units, code meanings, time axes, look-alike columns) and a list of **questions** for what the schema leaves open. It is told not to guess conventions — a guessed convention in the prompt is followed as confidently as a right one. A failed description does not fail the run.
- Chat reads the description and the conventions into the system prompt every turn. Exploration reads the conventions into its own prompt, so a proposal can carry the filter and the unit conversion.
- `ask.mjs` / `mappings.mjs --conventions` write the truth file’s `conventions` through the PATCH.

## Measured (wide corpus, fresh bases, #570 base)

What exploration wrote for `dwd_ord_dtl`, unprompted: grain (“one row is one line item within an order”), the keys, “money amounts … are in cents”, every status and channel code, six time axes, “amt_total is pre-discount, amt_pay is paid amount”, “price_old … no longer written after 2023 migration”. And eight questions, six of which are the truth file’s conventions verbatim in spirit: which `ord_st` count, which `pay_st` count, is `is_test` excluded, `amt_pay` or `amt_total` for revenue, are refunds already subtracted.

| | exploration covered | chat right |
|---|---|---|
| nothing | 0/18 | 2/18 |
| generated description only | 0/18 | 2/18 |
| description + the six conventions in the prompt | **3/18** | **11/18** (13 by SQL) |
| conventions + 18 definitions as a prose page, retrieval only (#520) | — | 14/18 |
| 27 confirmed definitions in the prompt (#520) | — | 17/18 |

Two things the table says. **The description alone moves nothing** — the model already gets grain and units from the schema document; what it lacked was never structure. **Conventions move exploration for the first time**: 0 to 3 definitions covered, and the five wrong proposals now all carry `ord_st IN (2,3,4) AND is_test != 1` and differ from the truth by a column choice (`amt_total` for `amt_pay`) or a filter variant (`ord_st` where money uses `pay_st`) — the same ambiguity the prose experiment found, now visible in proposals instead of answers. Six general rules in the prompt are worth 11; the prose page that also spelled out 18 definitions was worth 14; structured definitions 17. The order of work follows.

## Verified

- Unit: `parse_description` (fenced JSON, blank questions dropped, empty description refused).
- DB test: description and conventions do not overwrite each other across a re-run and a PATCH; a PATCH that omits conventions leaves them.
- The bench rows above, on a fresh base migrated by this branch.

Does not touch `Mappings.tsx` (#552). Touches `chat.rs` in the prompt-assembly block only (#548 rewrites the loop; the injection is twelve lines before the semantic-layer block).

### DIFF
--- a/crates/utopia-core/src/models.rs
+++ b/crates/utopia-core/src/models.rs
@@ -983,6 +983,14 @@ pub struct KnowledgeBase {
     /// 上次推完的时间。**答的是「上次看过没有」，不是「上次改过没有」**
     pub last_inference_at: Option<DateTime<Utc>>,
     pub ontology_lang: String,
+    /// 探索从 schema 写的数据描述：一行是什么、键、单位、码值、时间轴、相似列。
+    /// 只写 schema 说了的；每次探索重写
+    pub data_description: Option<String>,
+    /// 探索拿不准、需要库的主人答的问题（JSON 字符串数组）
+    pub data_questions: serde_json::Value,
+    /// 人写的约定：「测试单不算数」「有效订单是 2/3/4」这类 schema 里没有的规则。
+    /// 探索不碰它——量过：宽表语料上问数没有约定 2/18，有 14/18（#520）
+    pub data_conventions: Option<String>,
     pub created_at: DateTime<Utc>,
     pub updated_at: DateTime<Utc>,
 }
--- a/crates/utopia-server/src/api/chat.rs
+++ b/crates/utopia-server/src/api/chat.rs
@@ -625,6 +625,11 @@ pub async fn chat(
     )
     .await?;
     let workspace_id = kb.workspace_id;
+    // 数据描述（探索从 schema 写的）与约定（人写的）跟着进 system prompt。
+    // **每次都在，不靠检索碰运气**：约定写成一页文档只靠检索也到过 14/18，
+    // 但那是因为这批问题都在问指标才每题命中（#520）
+    let data_description = kb.data_description.clone().filter(|s| !s.trim().is_empty());
+    let data_conventions = kb.data_conventions.clone().filter(|s| !s.trim().is_empty());
 
     // 注册表在生成器之前取出来：下面那个 `async_stream!` 会把 `state` 整个搬走
     let live = state.live.clone();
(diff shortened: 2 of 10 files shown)


WaylandYang:
Review before merge. One addition: `PATCH /kbs/{id}` now records in the audit whether `data_conventions` was part of the change. The text goes into every query's prompt, so who changed it and when has to be answerable; the handler's authorization (Admin, same as name and description) was already right.

Noted, not changed: `describe_data` is a second full-schema model call serialised before the proposals call, and its failure is warned but not written to the run ledger. Worth a ledger column when exploration latency becomes a complaint; today the base is explored once.

This branch will be rebased onto dev after #567 lands (its base was renumbered to record 0036 under it).

PR #575, opened 2026-09-09
A question picks its definitions instead of taking the first thirty

### LINKED ISSUE #574: A question picks its definitions instead of taking the first thirty
Chat reads its confirmed definitions with `confirmed(kb, 30)`: the first thirty by concept name. The seeded upper bound on the wide corpus, 27 definitions, scored 17/18 (#520) — three short of the cap. A real base with a hundred definitions leaves seventy of them, the ones later in the alphabet, outside the prompt on every question. “Twenty-seven works” does not extrapolate.

The other wall is already visible at 27: the one miss in the seeded run was the model asking which of three adjacent names (gross profit, gross margin, gross merchandise value) the person meant. More definitions in the prompt means more of that.

## Shape

Definitions become retrievable, in the shape the ontology index already has: `concept_mappings.embedding` / `embedded_model` / `embedded_text` (migration `0048`), re-embedded when the model or the text changes, refreshed lazily before a lookup so nobody has to remember to.

Two channels, RRF-fused, top `k = 8` into the prompt:

- **vector** — “name: summary (unit)” embedded; the question embedded once; pgvector orders. Absent when no embedding model is configured.
- **lexical** — every confirmed definition pulled back and scored in Rust by how many of the question’s tokens it contains (ASCII words; CJK bigrams). A base has tens to a few hundred definitions; this is cheaper than Chinese tokenisation in SQL and needs no `pg_trgm`.

Degradation: vector + lexical → lexical only → the first `k` by name when nothing matches (today’s behaviour). No step is “no definitions”.

`GET /kbs/{id}/mappings/relevant?q=&k=` exposes the same lookup, so the bench can score it and a page can later show which definitions a question used.

## Measured by

**recall@k, without asking anything.** The bench already knows which definition each question needs (by number, #520), so `ask.mjs --recall 8` calls the endpoint per question and checks whether a correct row is in the top 8 — seconds per round, no chat model. Misses are annotated with the truth file’s `near` field: if they cluster on near-name pairs, that is the case for a reranker (a separate cut, with a third model slot and this same degradation); if they scatter, the embedded text is the thing to fix.

A base that exceeds the cap: `mappings.mjs --fresh --corpus wide --also tpch` mounts both sources into one base, seeding both truths gives 45 definitions across two sources. Then `ask.mjs` on each corpus checks that 17/18 survives selection.

WaylandYang:
Closes #574. Stacked on #571 (both touch the prompt block in `chat.rs`); retarget to `dev` once the stack lands.

Chat read its definitions with `confirmed(kb, 30)` — the first thirty by concept name. The seeded upper bound (27 definitions, 17/18 on wide) sat three short of that cap, and a base with a hundred definitions would leave seventy outside the prompt on every question.

## What changes

- **Definitions are retrievable.** `concept_mappings` gains `embedding` / `embedded_model` / `embedded_text` (migration `0048`), the ontology index’s shape: re-embedded when the model or the text changes, refreshed lazily before a lookup.
- **Two channels, RRF-fused, top 8 into the prompt** (`mapping_index`): vector (“name: summary (unit)”, pgvector) and lexical (every confirmed definition scored in Rust by the question’s tokens — ASCII words, CJK bigrams; no `pg_trgm`). Degradation: both → lexical only → the first `k` by name when nothing matches. Never “no definitions”.
- **`GET /kbs/{id}/mappings/relevant?q=&k=`** exposes the same lookup, so the bench can score it and a page can later show which definitions a question used.
- `ask.mjs --recall K` measures recall@k without asking anything (seconds); `mappings.mjs --also <corpus>` mounts a second source so a base can exceed the old cap.

## Measured

Two sources in one base, both truths seeded, **45 confirmed definitions**:

| | recall@8 | chat right | single-source seeded, for reference |
|---|---|---|---|
| tpch, 24 questions | 23/24 | **24/24** | 24/24 |
| wide, 18 questions | 13/18 | **15/18** | 17/18 |

Selection kept the tpch upper bound with 45 definitions in play (the old cap would have dropped fifteen of them). Wide lost two questions, and the recall misses are not near-name pairs: they are Chinese questions against English seeded definitions whose summaries are bench placeholders (`Paid orders — bench truth`). The lever is the embedded text, not a reranker — details on #574.

## Verified

- Unit: tokenizer (both languages, single ASCII characters dropped), lexical ranking (most shared tokens first, no-match empty).
- DB test: only changed or unembedded rows are re-embedded; vector search orders by distance and skips mismatched dimensions; `by_ids` keeps the requested order.
- The bench rows above on a base migrated by this branch; all 45 rows embedded on first lookup.

### DIFF
--- a/crates/utopia-server/src/api/chat.rs
+++ b/crates/utopia-server/src/api/chat.rs
@@ -580,11 +580,8 @@ pub async fn chat(
     //
     // 从前这里按 `confidence >= 0.75` 捞事实,而那个阈值是拿浮点数编码一个
     // 二值状态(提议 0.6 / 确认 1.0)。现在读 `status = confirmed`(0011)
-    let mappings = if mounted_sources.is_empty() {
-        Vec::new()
-    } else {
-        utopia_store::mappings::confirmed(&state.pool, kb_id, 30).await?
-    };
+    // 口径在下面按问题挑（`mapping_index::relevant`）——从前这里 `confirmed(kb, 30)`
+    // 按字典序取前三十条，一百条口径的库有七十条永远进不了提示词（#574）
     let can_write = utopia_store::access::kb_role(&state.pool, &user, &kb)
         .await?
         .is_some_and(|r| r >= Role::Editor);
@@ -600,6 +597,22 @@ pub async fn chat(
     if query.is_empty() {
         return Err(AppError::Validation("Missing user message".into()).into());
     }
+    // 语义层：跟这个问题有关的那几条确认口径进 system prompt——问数优先用确认口径，
+    // 而不是每次从 schema 猜。按问题挑而不是全塞：二十七条的上界 17/18 是在
+    // 三十条的上限之下量的，一百条口径靠字典序截断就不成立了（#574）
+    let mappings = if mounted_sources.is_empty() {
+        Vec::new()
+    } else {
+        crate::mapping_index::relevant(
+            &state,
+            kb_id,
+            kb.workspace_id,
+            &query,
+            crate::mapping_index::DEFINITIONS_IN_PROMPT,
+        )
+        .await
+        .map_err(AppError::Other)?
+    };
 
     // 会话持久化：有 id 则校验归属，无则以首句为题新建；用户消息即刻落库,
     // 上下文由服务端从库里拼——前端只送新消息
--- a/crates/utopia-server/src/api/mapping_routes.rs
+++ b/crates/utopia-server/src/api/mapping_routes.rs
@@ -154,6 +154,33 @@ pub async fn revisions(
     Ok(Json(json!({ "revisions": rows })))
 }
 
+#[derive(Deserialize)]
+pub struct RelevantQuery {
+    q: String,
+    k: Option<usize>,
+}
+
+/// 跟一个问题有关的口径，按相关度排——问数进提示词用的正是这一条检索（#574）。
+///
+/// 开成端点是为了两件事：测量台直接量 recall@k（那条对的口径在不在前 k 里），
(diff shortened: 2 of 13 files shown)


WaylandYang:
Re-measured after `afe152e6` (seeded definitions carry a real summary): wide recall@8 **18/18**, wide chat **18/18**; tpch recall 23/24 (the same paraphrase miss, answered correctly by chat). Table and reading on #574.

WaylandYang:
Review before merge. Two fixes:

- **The refresh had no bound on the request path.** `relevant` awaits `refresh` before the turn is answered, and `needing_embedding` returned every stale row, so the first question after a change of embedding model (or after this migration, on a base with many confirmed definitions) embedded all of them serially before speaking, and concurrent turns repeated the work. A turn now embeds at most 64 rows (`REFRESH_PER_TURN`); the rest are picked up by the following turns, and the lexical channel covers the gap meanwhile. The store function takes a limit, and the test asserts it.
- **The vector channel compared across models.** `vector_search` filtered on dimension only, so after a model switch to one of equal dimension, rows not yet re-embedded were ranked against the new model's query. It now takes the model and compares only rows embedded by it; a stale row falls back to the lexical channel instead of a false distance. Test case added.

Noted, not changed: the lexical channel matches tokens as substrings, so short English words ("is", "in") hit inside longer ones; it is noisy alone, but it only stands alone when no embedding model is configured, and RRF with the vector channel washes it out otherwise.

Not re-measured after these two changes. At 45 rows the bound does not bite and the base has one model, so the numbers on #574 should stand; I will re-run `--recall 8` on the merged dev before closing #574.

PR #590, opened 2026-09-10
An else branch that only holds an if is written as an else-if

lml2468:
**Corrected after checking CI — the original description of this PR had the motivation backwards. See the note below.**

`clippy::collapsible_else_if` fires on two blocks in `crates/utopia-store/src/temporal.rs` under clippy 1.93.0:

```
error: this `else { if .. }` block can be collapsed
  --> crates/utopia-store/src/temporal.rs:124:24
  --> crates/utopia-store/src/temporal.rs:142:24
```

It does **not** fire on the toolchain CI actually runs (rustc 1.98.1, 2026-09-01) — `#591`, which carries the unmodified file, passes `cargo clippy --workspace --all-targets -- -D warnings` on your runners. So this is not a CI fix, and nothing on `dev` is red because of it.

What's left is a small readability change that some clippy versions ask for: an `else` whose whole body is one `if` reads as an `else if`. `-18/+22`, behaviour unchanged.

Entirely reasonable to close this as cosmetic churn — no argument from me if you'd rather not take it.

Checked locally (macOS, rustc 1.93.0):

- `cargo fmt --all --check`
- `cargo test -p utopia-store` with `UTOPIA_DATABASE_URL` set — 178 passed, 0 failed

### DIFF
--- a/crates/utopia-store/src/temporal.rs
+++ b/crates/utopia-store/src/temporal.rs
@@ -121,17 +121,15 @@ pub async fn reconcile_new_fact(
                 if new_confidence < AUTO_CLOSE_MIN_CONFIDENCE {
                     record_conflict(pool, kb_id, old.id, new_fact_id, "low_confidence").await?;
                     report.conflicts += 1;
-                } else {
-                    if let Some(id) = close_superseded(
-                        pool,
-                        new_fact_id,
-                        of,
-                        old.valid_from_precision.as_deref().unwrap_or("day"),
-                    )
-                    .await?
-                    {
-                        report.corrected.push(id);
-                    }
+                } else if let Some(id) = close_superseded(
+                    pool,
+                    new_fact_id,
+                    of,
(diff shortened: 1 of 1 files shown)


lml2468:
Correcting my own PR description before anyone spends review time on it.

I claimed this lint would turn CI red once runners picked up a newer stable. That's backwards. Your runners are already on rustc 1.98.1 and clippy there does **not** flag these blocks — #591 carries the unmodified `temporal.rs` and its Clippy step is green. The lint fired on my local 1.93.0, which is nine months behind.

So there is no CI breakage here, and the only thing this PR offers is the collapse itself. I've rewritten the description to say that. Happy to close it if it isn't worth the churn.

lml2468:
Closing this — as noted above, it isn't a CI fix: clippy on the toolchain your CI runs (1.98.1) doesn't flag these blocks, and the collapse on its own isn't worth the review time.

Leaving it here in case the lint comes back on a future stable. Sorry for the noise.

PR #591, opened 2026-09-10
A script whose usage line says ./ is executable

lml2468:
Both shell scripts under `scripts/` carry a shebang and document themselves as being run directly:

```
# 用法: ./scripts/smoke.sh [BASE_URL]  (默认 http://localhost:1516)
# 用法: ./scripts/e2e_type_drift.sh
```

Neither is committed with the executable bit, so following those lines gives:

```
$ ./scripts/smoke.sh
zsh: permission denied: ./scripts/smoke.sh
```

You have to know to type `bash scripts/smoke.sh` instead — a small stumble, but it lands on someone verifying a fresh deployment, which is exactly when you want the documented command to just work.

Mode change only, `100644 → 100755`, no content touched. `e2e_type_drift.sh` is included because it has the identical problem; say the word and I'll split it out.

Verified after the change: `./scripts/smoke.sh` runs and passes against a local `docker compose --profile app` deployment.


WaylandYang (approved):
Both scripts carry a bash shebang and a usage line that says `./`; a mode change is the whole fix, and these are the only two shell scripts under `scripts/`, so nothing is left behind. Thanks for checking it against a live deployment.

PR #583, opened 2026-09-10
A fact quotes the words that name each side

WaylandYang:
Closes #582.

## What was wrong

#559, #578 and #581 are one failure with three faces: the model attaches a fact to the wrong participant. "Former OpenAI personnel have founded … Anthropic" became `OpenAI —founded→ Anthropic`; "companies using OpenAI contacted Anthropic" became `OpenAI —contacted→ Anthropic`. Each face got a rule in the prompt and a guard with a word list. The guards measured that the rules get about half compliance, and a word list only knows the shapes it has already seen: #579's list caught `former X personnel` and `X's investors`, mistook `partners` for a group noun, and would have let `the company's board` through.

## What changed

**The model quotes; the machine judges.** Two fields on every fact, `subject_span` and `object_span`: the verbatim words in `quote` that name each side. Copying is a task the model does reliably; deciding whether a description "counts" as the entity is the task it does badly. Rule 8d shrinks from a paragraph with two JSON examples (732 characters) to one sentence about copying the spans (340).

At ingest, `verify_span` compares each span with the name of the entity the fact is **bound to** (the entity its `subject_ref` / `object_ref` points at, or the written name when there is no reference) and sorts it into a tier, no model call:

1. **In the quote?** (case and whitespace aside). Otherwise the fact is handled as if it had no span and `span_not_in_quote` is recorded, so the rate reads per model.
2. **The bound name itself**: the same recall keys as resolution (same name, same stem, generic suffix), or the span's words are all in the name (`Altman` for `Sam Altman`, `Anthropic` for `Anthropic, PBC`), after stripping punctuation, possessives and a leading article. → kept.
3. **Another declared entity's name** (exact, or the span's words all inside it: `OpenAI's board` → `OpenAI's board of directors`) → the fact is **rebound** to it (`span_rebound`).
4. **The name with words after it, or with a possessive** (`Former OpenAI personnel`, `The Verge reporter`, `Anthropic's safeguards`) → the name is a modifier inside a larger phrase: a **description**. A described subject drops the fact (`subject_described`); a described object keeps the fact with the span as its literal value (`object_described`), the path #568 uses for undeclared objects. Words after the name that all look like proper names (`Anthropic PBC`, `OpenAI Global, LLC`) are the same thing and are kept.
[...]

Issue #582, opened 2026-09-10
A fact quotes the words that name each side, and the check is mechanical


WaylandYang:
#559, #578 and #581 are one failure with three faces: the model attaches a fact to the wrong participant. It writes `OpenAI` for "former OpenAI personnel", mints a node for "lawsuit against OpenAI", keeps "Vice President of Research" and drops the OpenAI it belonged to. Each got a rule in the prompt and a guard with a word list; the guards measured that the rules get about half compliance, and the word lists only know the shapes we have already seen.

The general fix is to stop asking the model to obey a rule and ask it to **quote**. A fact already carries `quote`, the sentence it rests on. Add two fields: `subject_span` and `object_span`, the verbatim words in the quote that name each side. Copying is a task models do reliably; deciding whether a description "counts" as the entity is the task they do badly. Then the check is mechanical and needs no model call:

1. **The span must be in the quote.** Otherwise the fact is treated as if it had no span (`span_not_in_quote` signal, so the rate reads per model).
2. **The span must name the entity it is bound to.** Same name after normalisation, or the same stem (`Acme` / `Acme Corporation`), or an alias. When it does not:
   - the span names another entity declared in this response or known to the base → rebind to that one (`span_rebound`);
   - the span names nothing → it is a description. A described subject drops the fact (`subject_described`); a described object keeps the fact and stores the span as the literal value (`object_described`), as #568 does for undeclared objects.

"Former OpenAI personnel" fails check 2 against `OpenAI` and matches nothing → dropped, with the phrase in the ledger. "companies using OpenAI" likewise. "the company's board" or "he" as a subject span: the same path, with no word list at all. The #579 guard and its list of group nouns retire.

What this does not cover: a fact the model never wrote (#581). That is a recall problem and stays separate.

## Cost

Two short fields per fact in the output, no extra call. The prompt loses rule 8d's paragraph and example; the rule becomes one sentence about the spans.

## Acceptance

Re-extract `openai.txt`, `removal-of-sam-altman-from-openai.txt` and `anthropic.txt` and compare with the #579 run of the same three (604 entities, 1,288 facts, no OpenAI–Anthropic edge from either sentence):

- the two sentences still yield no OpenAI–Anthropic edge, now caught by the span check rather than the word list;
- `span_not_in_quote` rate per fact, and what the model does with the fields (are they filled, are they verbatim);
- `subject_described` / `object_described` examples read as descriptions; `span_rebound` examples read as correct rebinds;
[...]

PR #565, opened 2026-09-09
The graph answers with paths, neighbours and a timeline

WaylandYang:
Closes #558.

## What changed

The model's graph tools were a name lookup and a fact dump. Now the query work happens on the server:

- **`paths_between(from, to, max_hops ≤ 3, at?, as_of?)`** (`utopia-store::paths`): the chains of facts joining two entities, shortest first, at most ten, deduplicated (the same edge is often two facts, one dated and one not). The search meets in the middle (a's two hops against b's one), never expands a hub past the first hop, and runs both clocks: with `at` every edge on a path must hold at that moment, with `as_of` the edges and the entity ownership are read as the base held them then.
- **`neighbors(entity, predicate?, object_type?, at?)`**: one hop, grouped by predicate, each neighbour with its type and range.
- **`timeline(entity, since?, until?, predicate?)`**: the entity's facts that state a world-time position, in order; undated facts are counted and left out.
- **`entity_facts`** takes `predicate`, `object_type`, `since` / `until` and `limit`, and its output is grouped by predicate with a header that says how much was cut and where to narrow. A filter that matches nothing lists the predicates the entity does have; the model's guess ("board member") and the base's word (`comprised`) rarely coincide (#560).
- **`find_entities`** ranks typed before untyped, exact before partial, heavy before light; untyped phrase entities are left out when anything typed matches (#559); the reply names the best match when one dominates.
- **Every entity parameter takes a name as well as an id.** A name is resolved by the same ranking (canonical name or alias, substring, case-insensitive), and the reply's first line says which entity was chosen and which others matched, with their ids. That saves a `find_entities` round and the "which OpenAI?" that usually followed it.
- **Rendering:** a fact with no stated start reads `undated` (it read `attested 2026-09-06T12:23:58.175078Z`, and Qwen quoted that as the date something happened); one the text says is over reads `ended, date unknown`.
- MCP exposes the three new tools; the manual's MCP page lists them; the chat shows them with the entity and facts icons (six lines in `Chat.tsx`).

## As the model sees it

`paths_between("OpenAI", "Anthropic")`:

```
"OpenAI" = OpenAI (ResearchProject, 433 facts); other matches: OpenAI's board of directors (Organization, 16 facts, 01a076ad-…); OpenAI Global, LLC (Corporation, 10 facts, …); … Pass an id to choose another.
"Anthropic" = Anthropic (Organization, 151 facts); other matches: Anthropic Institute (ResearchOrganization, 2 facts, …). Pass an id to choose another.
[...]

Issue #514, opened 2026-09-08
Type resolution reads the whole entity table once per subject


roycegeo:
`resolve_batch` walks its subjects one at a time, and each iteration goes to the database twice:

```rust
// crates/utopia-server/src/type_resolution.rs:218
for (i, s) in subjects.iter().enumerate() {
    let descendants = match s.coarse_id {
        Some(c) => utopia_store::resolution::descendants_of(&state.pool, kb_id, c).await?...,
        None => HashSet::new(),
    };
    ...
    let raw = utopia_store::resolution::nearest_typed_entities(&state.pool, kb_id, s.id, NEIGHBOURS).await?;
```

`BATCH` is 60 and `MAX_ROUNDS` is 10, so a resolution run is up to 600 iterations of this, taken in order.

`nearest_typed_entities` is not a cheap query. It reads the whole entity table:

```sql
-- crates/utopia-store/src/resolution.rs:2744
ORDER BY e.profile_embedding <=> (SELECT v FROM me)
LIMIT $3
```

`entities.profile_embedding` has no index either, so each call computes cosine distance against every typed, unmerged entity in the knowledge base. Sixty subjects means sixty full passes over the same table, back to back, per round.

## The cheap half

`descendants_of` is keyed on `s.coarse_id` and nothing else. The coarse class comes from extraction and is drawn from a small vocabulary — person, organization, product and a handful more — so across sixty subjects the same recursive CTE is issued over and over for the same few arguments. A `HashMap<Uuid, HashSet<Uuid>>` held for the batch removes almost all of them and cannot change an answer: same input, same query, same result, within one batch where the ontology is not moving.

This is the part to do first. It is small, it is provably behaviour-preserving, and it is measurable on its own.

## The expensive half

The sixty `nearest_typed_entities` calls are independent of each other — nothing in iteration *i* reads anything iteration *i-1* wrote. They are serial only because the loop is.

The pattern is already established twenty lines up. [`type_resolution.rs:184`](crates/utopia-server/src/type_resolution.rs#L184) runs the two ontology retrieval routes concurrently:

```rust
let (profile_hits, name_hits) = tokio::join!(...);
```

So the awareness is there; the per-subject loop just did not get it. Bounded concurrency over the retrieval half of the loop — gather the neighbours for all sixty subjects, then run the per-subject reasoning against the gathered results — turns sixty serial scans into a handful of concurrent ones.

[...]

PR #594, opened 2026-09-10
dev → main: extraction that quotes its evidence, graph paths in chat, rules that compute, the ontology as tables

WaylandYang:
Sixty-three commits on dev since #478 (plus the #479 back-merge). Merge commit, as always for dev → main; DCO is red on this PR because it aggregates earlier commits, and the required checks are backend, web and migrations.

## Extraction that checks itself
- A fact quotes the words that name each side, and the machine checks them against the entity the fact is bound to; the #579 word list retires (#583, measured in #582 and on the recall bench: 50/52, level with the build before it).
- A group described by its relation to X is not X (#579); an undeclared name does not become a node (#568); one relation, one predicate, with schema.org's `supersededBy` honoured at import and new relations folded by meaning (#569); a predicate is named one way (#536).
- A written quantity is a value, not a node, and a quantity is never an entity (#586, #587); a relation says whether it happens or holds (#593); an event holds at the moment it names (#487); a derivation rewinds with the second clock (#557).
- A Chinese name does not stop the queue, and a stated figure is written down (#523); a new base starts with no ontology pack (#585).

## The graph answers
- The chat answers with paths, neighbours and a timeline: `paths_between` up to three hops on both time axes, namesakes ranked by the question, predicates aligned by meaning (#565); a promise to search is not an answer (#543).
- A question picks its definitions instead of taking the first thirty (#575); a definition can be written by hand (#563); the schema document is searched, not extracted (#561); the semantic layer gets measured, and exploration learns to align and to say what it cannot tell (#567, #571); an action is a declared call, written down (#531).

## Rules and governance
- A rule's criterion is a column; a rule may say *or*, once; *not one of* is not silently skipped; a rule reads what a rule concluded and computes what it concludes, written down; a rule is an edge on the schema diagram and is written in a dialog (#490, #484, #494, #485, #493, #495, #483, #482).
- The adjudicator looks before it asks (#475); an automatic merge is gated by what it can undo (#474); a decision records why (#473); record 0028 becomes 0031 (#489).

## Faster where it was measured
- A vector index is built by a job, expression HNSW with the dimension a literal (#566, closing #512 and #514); ingest embeds four batches at a time and both ingests share the loop (#539); a search does not hold a runtime worker (#535); a purge judges its blobs in one question (#537); idle workers wake when jobs are queued (#521).
- Model probes report reachability and authentication (#541); LLM errors preserve their response bodies, and the client's literals read at the point of use (#538, #542); an unset variable stays unset on the way into the container (#572); bulk mapping decisions (#552, from #505); RSS hydration summaries scoped to each source (#462); the lakehouse form keeps its one variable and does not blank the page (#576, #577).

## The interface
- The ontology reads as three tables and its tables read like a table; class hierarchy links show; selecting a class goes straight to its highlight, drawn last (#524, #540, #544, #534, #584); the schema diagram draws the biggest classes, on one canvas drawn once (#511, #499).
- Classes, properties, attributes and knowledge edit in dialogs while the panel shows (#480); the docked panel and the table say where you are (#545); a floating panel does not borrow the colour behind it, a glass panel does not flash white (#525, #532); a page renders before its data arrives and does not refetch what the stream keeps fresh (#529, #533); a conversation reads at reading size (#551); a fact row leads with the fact (#506); re-running a document is an icon (#491); KB members read like the deployment's user list (#492); the import history shows what each import did (#481).


WaylandYang:
Closing this release PR: dev has moved a long way past the branch point since it was opened (the shadcn interface, the light-theme fixes, the ledger and MCP work). A fresh dev → main PR will be opened when the next release is cut, so that its body describes what is actually in it.

Issue #564, opened 2026-09-09
Expose conflict and review state in the supported RDF read contract


ansonnmm:
## Summary

Following the clarification in #550 that RDF export is Utopia's supported machine-readable integration boundary, there is one remaining gap that seems important for downstream readers:

> the export contains the assertions themselves, but not the conflict / review state Utopia has already established about those assertions.

This means an external system can recover two evidence-backed facts, their provenance, validity and history, but cannot reliably tell from the supported contract that Utopia considers them currently contested.

## Context

As clarified in #550:

- RDF export is the supported machine-readable read contract;
- MCP is the agent-facing surface;
- `/api/v1` is UI-internal and is not a compatibility promise.

The RDF export already preserves the important epistemic lineage:

```text
entity
→ assertion
→ world time
→ record time
→ correction / supersession
→ evidence quote
→ source document
→ derivation / rule / premises
```

The missing piece is conflict and review state.

## Example

Suppose the knowledge base contains two evidence-backed assertions such as:

```text
Acme revenue = 100M
Acme revenue = 130M
```

and Utopia has already detected that they form a conflict.

Internally, Utopia can preserve both assertions and surface the conflict for review.

A downstream reader of the RDF export currently sees both assertions, but does not receive the epistemic fact that:

```text
these assertions are contested
```

Without that state, the integration has two undesirable options:

1. treat both assertions as ordinary independent facts; or
2. rebuild Utopia's conflict-detection logic downstream.

The second would duplicate logic that Utopia already owns, while the first loses part of Utopia's epistemic model.

## Desired property

The supported machine-readable contract should make it possible to determine that an exported assertion participates in an unresolved or otherwise relevant conflict/review state.

I am not suggesting a particular RDF vocabulary or storage model.

The minimum useful information seems to be equivalent to:

```text
Conflict {
  id
  kind
  status
  affected_fact_ids[]
}
```

or RDF relationships carrying the same semantics.

The important invariants are:

- the conflict has a stable machine-addressable identity, if Utopia treats it as an addressable object;
- affected assertions can be joined to it through their existing stable IRIs/UUID-derived identities;
- conflict kind is machine-readable;
- current resolution/review status is machine-readable;
- a downstream reader does not need to reconstruct conflict detection from the exported assertions.

## Scope

This issue is about exporting **Utopia's existing epistemic judgment** that facts are contested.

[...]

PR #601, opened 2026-09-10
Expose structured MCP reads with stable ledger identities

### LINKED ISSUE #550: Clarify or expose a stable structured read contract for facts, evidence, proofs, history, and conflicts
## Summary

Utopia already stores and exposes a rich epistemic chain internally and through several structured HTTP endpoints:

```text
entity
→ asserted fact
→ evidence / quote
→ source document
→ validity
→ history / correction
→ derived fact
→ proof
→ conflict / review state
```

However, the documented MCP read surface currently returns much of this information as formatted human-readable text rather than identity-rich structured data.

I would like to clarify what Utopia considers its supported external machine-readable contract for this information.

## Tested version

Tested against stock Utopia at:

```text
b03a0f20a3d053323a3ec2669c7c9b4aeb9a8e14
```

No source, schema, or migration modifications were made.

## What I verified

Using the existing HTTP and MCP interfaces, I was able to verify that Utopia has enough underlying information to reconstruct:

- entity identity;
- asserted fact identity and world-time validity;
- evidence quote and chunk identity;
- source document identity and version;
- correction / history events;
- derived fact identity;
- ordered proof premises;
- conflicts / review items.

Most of this information is already available through structured `/api/v1/...` HTTP routes.

The uncertainty is whether those routes are intended to be stable external integration contracts.

## Current MCP limitation

The documented MCP interface is clearly intended for external clients, but tools such as `entity_facts` primarily return formatted text.

For example, the result can describe validity and identify a derived result with something similar to:

```text
[rule: ...]
```

but the external caller does not receive a machine-composable structure containing identities such as:

```text
fact_id
evidence_id
document_id
derived_fact_id
proof / premise ids
conflict identity / status
```

Similarly, change/history output contains useful human-readable information, but does not expose enough stable identities to reliably join it back to the fact/evidence/proof chain.

This makes the MCP surface useful for an interactive agent, but difficult to use as a durable machine-to-machine knowledge interface without parsing formatted prose.

## Questions for maintainers

Could you clarify which of the following is the intended direction?

### 1. Existing structured `/api/v1` routes are public external contracts

If so:

- Are integrations expected to depend on the entity, fact, evidence, history, proof, graph, and review routes?
- Is there a compatibility or deprecation policy for those response shapes?
- Are opaque entity/fact/evidence/derived IDs intended to remain usable across versions?

### 2. MCP is the intended public integration contract

If so, would you consider exposing structured content alongside the human-readable text, including stable opaque identities and lineage relationships?

### 3. Another external API layer is planned

If neither of the above is intended as the durable contract, is there a planned structured read API for this use case?

## Minimal information that seems sufficient

A useful structured external contract would not need to expose database rows or internal implementation details.

Opaque IDs plus typed relationships would be sufficient. For example:

```text
EntityRef {
  id
  name
  type
}

AssertedFact {
  id
  subject
  predicate
  object / value
  valid_from
  valid_to
}

EvidenceRef {
  id
  fact_id
  quote
  chunk_id
  document_id
}

DerivedFact {
  id
  subject
  predicate
  object / value
  rule
}

Proof {
  derived_fact_id
  premises[]
}

HistoryEvent {
  id
  timestamp
  entity_id
  fact_id?
  event_type
}

Conflict {
  id
  fact_ids[]
  kind
  status
}
```

The exact schema above is only illustrative.

The important properties are:

- stable opaque identity;
- machine-composable lineage;
- asserted and derived facts remain distinguishable;
- evidence and source provenance remain recoverable;
- history/corrections can be joined back to the affected facts;
- conflicts can be addressed as structured objects rather than reconstructed from text.

## Why this may be useful beyond one integration

One of Utopia's differentiating properties is that it preserves more than the current graph state.

It also preserves:

- evidence;
- provenance;
- temporal validity;
- changing belief/history;
- derivations and proof chains;
- conflicts;
- auditability.

A structured external read contract would allow downstream agents and applications to preserve those guarantees instead of flattening the result into text and then attempting to reconstruct structure from prose.

It would also provide a clean boundary between Utopia's internal schema—which can continue to evolve—and external integrations that should not depend directly on PostgreSQL tables, migrations, or Rust implementation details.

## Related documentation

Current MCP documentation at the tested commit:

https://github.com/deeplethe/utopia/blob/b03a0f20a3d053323a3ec2669c7c9b4aeb9a8e14/web/src/docs/mcp.md#L20-L52

## Request

The immediate request is primarily clarification:

> What external structured read surface, if any, should integrations treat as a supported compatibility boundary?

If MCP is intended to be that boundary, structured identity-rich responses for the epistemic chain above would close the current gap.

If the existing structured HTTP routes are already intended to be that boundary, documenting their compatibility expectations would also resolve the uncertainty.

Floating-Y:
MCP readers currently have to parse prose to recover facts and provenance.
Add structuredContent alongside existing text for find_entities,
search_chunks, entity_facts, and changes, using the same UUIDs as RDF exports.

- Preserve fact identities, both time axes, source document IDs, raw values,
  relation qualifiers, and explicit limit metadata.
- Return read failures as MCP errors instead of successful empty collections.
- Document the supported structured read contract and its limitations.
- Restore RDF links to derived premises in Turtle and JSON-LD.
- Add authenticated MCP integration tests and PostgreSQL CI coverage.

No database migration or new dependency is required.
Conflict/review export (#564), chunk provenance in RDF, and SPARQL
remain separate work.

## Validation

- Formatting and workspace Clippy passed.
- Workspace tests with PostgreSQL required: 604 passed, 1 ignored.
- Frontend tests: 18 passed.
- Frontend production build passed.

Closes #550.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,12 @@ jobs:
           failed=$(grep -o '[0-9]* failed' store-tests.log | awk '{s+=$1} END {print s+0}')
           echo "utopia-store against Postgres: **${passed} passed**, ${failed} failed — a missing database fails this job instead of skipping" >> "$GITHUB_STEP_SUMMARY"
 
+      - name: MCP structured reads against Postgres
+        run: cargo test -p utopia-server api::mcp::tests
+        env:
+          UTOPIA_DATABASE_URL: postgres://utopia:utopia@localhost:5432/utopia
+          UTOPIA_TEST_REQUIRE_DB: "1"
+
   web:
     runs-on: ubuntu-latest
     defaults:
--- a/crates/utopia-core/src/models.rs
+++ b/crates/utopia-core/src/models.rs
@@ -709,6 +709,10 @@ pub struct GraphEdge {
 #[derive(Debug, Clone, Serialize, sqlx::FromRow)]
 pub struct EntityFact {
     pub id: Uuid,
+    pub recorded_at: DateTime<Utc>,
+    pub invalidated_at: Option<DateTime<Utc>>,
+    pub supersedes: Option<Uuid>,
+    pub document_ids: Vec<Uuid>,
     /// out = 该实体为主语；in = 为宾语
     pub direction: String,
     /// 本体没认下这条关系时回落到原文说法；两者都拿不出时为 None（更早的历史数据长这样）
@@ -1217,6 +1221,13 @@ pub struct OntologyDefect {
 #[derive(Debug, Clone, Serialize, sqlx::FromRow)]
 pub struct DerivedFactView {
     pub id: Uuid,
+    pub predicate_id: Uuid,
+    pub object_value: Option<serde_json::Value>,
+    pub rule_id: Option<Uuid>,
+    pub attribute_rule_id: Option<Uuid>,
+    pub invalidated_at: Option<DateTime<Utc>>,
+    pub valid_from_precision: Option<String>,
+    pub valid_to_precision: Option<String>,
     pub subject_id: Uuid,
     pub subject: String,
     /// 字面值结论（业务规则的归类与属性）没有实体宾语（0021）
--- a/crates/utopia-server/src/api/chat.rs
+++ b/crates/utopia-server/src/api/chat.rs
@@ -199,7 +199,7 @@ pub(super) fn tools_schema(can_write: bool, data_source_names: &[String]) -> ser
 ///
 /// 判据直接取自工具表里的 `required`：加一个必填参数，这里自动跟上，
 /// 不必记得来改第二处。
-fn check_call(
+pub(super) fn check_call(
     tools: &serde_json::Value,
     name: &str,
     raw_args: &str,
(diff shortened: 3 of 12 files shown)


WaylandYang:
Reviewed against `dev` after #600/#602 — no logical conflict with the qualifier work, identities are the real row ids, invalidated rows behave as the export does, and the tests read every id back. Thank you; merging.

Three small things I'll take as a follow-up rather than hold this on (tracked in an issue):

1. `tools_graph.rs` `resolve()` maps a database error to `"could not look up entities"`, which `entity_facts` then wraps as `Invalid entity: … (expected a name, or the uuid returned by find_entities)` and `neighbors` / `timeline` / `paths_between` report as `Unknown entity` with `isError:false`. An outage reads as a bad argument; a distinct "could not read" result, as `find_entities` already returns, is the fix.
2. `docs/decisions/0020` gets an inserted paragraph without a date; the records README asks for a dated revision note in place.
3. `web/src/docs/mcp.md` tabulates every `structuredContent` field under a heading called "contract" but never says whether a removal or rename needs a record. One sentence either way.

Also noted: refused calls (the shared `check_call` guard) skip the `mcp.tool_called` audit row; `search_chunks.limit_reached` and an entity-valued qualifier are untested.

PR #611, opened 2026-09-11
fix(ingest): strip NUL bytes from PDF text before chunk insert

plpycoin:
## Problem

`utopia_ingest::parse` on PDFs occasionally forwards NUL bytes from font
ToUnicode / encoded content streams into `parsed.text`.
`utopia_ingest::chunk_text` does not sanitize, and
`utopia_store::documents::replace_chunks` binds each chunk's text straight
into the `chunks.text` TEXT column. The INSERT then fails with
`invalid byte sequence for encoding "UTF8": 0x00`, the transaction rolls
back, and the document is pinned as `failed`.

Repro: upload `compbiobench.pdf` (sha256 `e0b23decafd1545391b5a255a832a096f57a08d00f7922c091d471ff2c910485`)
to any KB on a workspace whose embedding is configured. Document error:
`error returned from database: invalid byte sequence for encoding "UTF8": 0x00`.

## Fix

Strip NUL at the earliest point `pipeline.rs` holds the text, so downstream
chunking / indexing / embedding all see clean input. One file,
`crates/utopia-server/src/pipeline.rs`:

```rust
let parsed_text: String = parsed.text.chars().filter(|c| *c != '\0').collect();
let text_len = parsed_text.chars().count() as i32;
...
let pieces = utopia_ingest::chunk_text(&parsed_text);
```

## Why at `pipeline.rs`

- The byte originates in `utopia_ingest::parse` — could be fixed there too,
  but then it only fixes PDFs; HTML / docx may carry other control bytes
  through the same downstream path.
- Cleaning at the `pipeline.rs` boundary covers every parser uniformly
  without touching `utopia-ingest` (which is also consumed by tests and
  CLIs and has no direct DB write dependency to motivate the change).
- Costs one extra `chars().filter(...).collect()` per document — trivial
  relative to Tantivy reindex + embedding roundtrip.

## Verification

- `cargo check -p utopia-server` clean.
- `cargo build -p utopia-server` clean.
- Reprocess of `01a08f6d-ccd1-77a2-9004-16ca7874a9fb` once the new image is
  rolled out should move the document from `failed` to `ready`.
  (Live dev box is running a pre-fix container so this PR cannot be
  end-to-end verified there without redeploy — flagging for reviewer.)

### DIFF
--- a/crates/utopia-server/src/pipeline.rs
+++ b/crates/utopia-server/src/pipeline.rs
@@ -62,10 +62,13 @@ async fn run(state: &AppState, document_id: Uuid) -> anyhow::Result<()> {
     let filename = doc.filename.clone();
     let parsed =
         tokio::task::spawn_blocking(move || utopia_ingest::parse(&filename, &bytes)).await??;
-    let text_len = parsed.text.chars().count() as i32;
+    // PDF 提取偶尔会在流里夹 NUL（字体 ToUnicode / 编码内容），Postgres TEXT 不收
+    // 0x00，落到 chunks.text 直接报 invalid byte sequence，整篇 failed。这里先剥掉。
+    let parsed_text: String = parsed.text.chars().filter(|c| *c != '\0').collect();
+    let text_len = parsed_text.chars().count() as i32;
 
     // 2. 分块 + 入库
-    let pieces = utopia_ingest::chunk_text(&parsed.text);
+    let pieces = utopia_ingest::chunk_text(&parsed_text);
     let chunk_pairs =
         utopia_store::documents::replace_chunks(&state.pool, doc.kb_id, document_id, &pieces)
             .await?;


WaylandYang:
Thanks for the careful write-up — the diagnosis and the placement both hold up. I checked the surrounding code independently before reviewing:

- `utopia_store::documents::replace_chunks` has exactly one production call site, `pipeline.rs:70`. This really is the only path into `chunks.text`.
- `utopia_ingest::parse` has exactly one production call site, `pipeline.rs:64`; the other two are inside `#[test]` blocks in `rss_full_content.rs`. So cleaning at this boundary does cover every parser, as you argue.
- `parsed.` is referenced exactly twice in `pipeline.rs`, and this PR changes both, so no raw text leaks further downstream.
- NUL is the only byte Postgres will reject here: `TEXT` refuses `0x00`, and a Rust `String` is already guaranteed valid UTF-8, so nothing else malformed can reach the bind. Filtering `\0` is sufficient, not just necessary.

Three things before this can go in.

**1. DCO.** Every commit in this repo needs a `Signed-off-by` line; the check is currently `ACTION_REQUIRED`. `git commit --amend -s && git push --force-with-lease` will clear it.

**2. The copy is unconditional.** `chars().filter(...).collect()` allocates a second copy of the entire document text for every document, while a NUL byte is the rare case. The equivalent that does not pay for the common path:

```rust
let parsed_text = if parsed.text.contains('\0') {
    parsed.text.replace('\0', "")
} else {
    parsed.text
};
```

`parsed` is not used after this point, so the clean path moves instead of copying, and `replace` avoids walking the string char by char. Your note that the cost is small next to the Tantivy reindex and the embedding roundtrip is true, but that is an argument for tolerating an allocation, not for making one.

**3. It is not testable where it sits.** `pipeline.rs::run` needs an `AppState` and a live database, so there is no way to pin this behaviour down. Pulling the strip into a small named function — either a private `fn strip_nul(text: String) -> String` in `pipeline.rs`, or next to `chunk_text` in `utopia-ingest` — makes it a three-line unit test, and the next refactor cannot quietly drop it.

One thing worth adding to the PR body rather than changing: `text_len` now counts characters after stripping. That is an improvement, since the stored length matches what was actually stored, but it is a semantic change and a future reader should not have to infer it from the diff.

On verification: you flagged that the dev box runs a pre-fix container. That part can be covered on our side — a PDF carrying a NUL through a build with this patch, confirming the document lands `ready` instead of `failed`. Happy to run it once the three points above are addressed.

The branch is also behind `dev`, so the backend, migrations and web checks have not run yet; updating it will kick them off.

WaylandYang:
Thanks — the fix itself reads right. A PDF's text layer can carry `0x00` (font `ToUnicode` maps and encoded content streams are the usual sources), Postgres `TEXT` will not take it, and the insert into `chunks.text` fails the whole document rather than the one byte. Stripping before chunking is the right place: it keeps `text_len` and the chunk offsets consistent with what is actually stored.

The only thing blocking it is DCO — the commit has no `Signed-off-by` line, so the check stays red and we cannot merge it. Every commit in this repository needs one, matching the commit author.

To fix it on the single commit you have:

```bash
git commit --amend -s --no-edit
git push --force-with-lease
```

That adds `Signed-off-by: Your Name <your@email>` to the message and re-runs the check. Nothing else needs to change.

One question before we merge, and it does not need a code change if the answer is "no": should the strip be counted anywhere? Right now a document with NUL bytes goes from "extraction failed" to "extraction succeeded" with no trace that anything was dropped. We are not asking you to add that here — it is a separate decision about whether silent repair should leave a mark — but if you already know how often this shows up in your corpus, that number would be useful to us.

WaylandYang:
Picked this up in #630 so the fix can ship without waiting on the sign-off — you're credited as co-author on the commit, and the fix is yours.

Two things added on top: a test that runs a document with NUL bytes through the real `process_document` against Postgres (with your strip reverted it fails with the exact `0x00` error from your report), and the strip now borrows instead of copying when a document has nothing to remove, which is almost every document.

Thanks for the precise report — the sha256 and the exact database error made this quick to confirm. Closing in favour of #630.

Issue #610, opened 2026-09-11
An upload can say when it was written


WaylandYang:
Found while rehearsing the litigation demo corpus (16 dated announcements).

A file that comes in through the multipart upload (`POST /kbs/{id}/documents`) gets `doc_time = now()` and `doc_time_source = 'upload_time'`. The only ways to give a document its own date are the JSON ingest routes (`/kbs/{id}/ingest`, `/sources/{id}/ingest` with `doc_time`). Announcements, filings and minutes carry their date in the title or the first line, and the ledger's attested anchors (0022) — "known from which document, dated when" — read `doc_time`; with upload time every anchor is today, a dateless mention cannot find the span it belongs to, and the timeline's "known since" is meaningless for anything uploaded in bulk.

Two cuts worth considering, either or both:

- [ ] The multipart upload accepts a `doc_time` field (per request, or per file via a JSON sidecar field), stored with `doc_time_source = 'upload'`.
- [ ] A conservative detection for text-like uploads: a full date in the first line or a leading "（YYYY年M月D日）" / ISO date in the title sets `doc_time` with `doc_time_source = 'content'`; anything less than a full date leaves upload time. The panel shows the source so a person can tell which it was.

The rehearsal worked around it by pushing each file through `/kbs/{id}/ingest` with the date parsed from the title.

Floating-Y:
## Remaining scope after #645

#645 recognizes a complete, standalone opening dateline in text uploads. I suggest finishing #610 with:

- Recognize a valid full ISO or Chinese date at the start of each document's title or heading. Keep the upload-time fallback when the date is invalid or ambiguous.
- Show `doc_time` and `doc_time_source` in the document panel, separately from the upload timestamp.
- Verify multi-file uploads and the resulting evidence anchors and timeline with database-backed tests.

This follows the per-document content approach established in #645.

Could you confirm whether “title” includes the uploaded filename, an in-document heading, or both? If the original 16-document corpus includes PDF or Office files, should their date extraction be required to close #610 or tracked separately?

rollroyces:
Picking this up. Plan, mirroring #645's text-only cut:

**Scope of this PR:** text uploads (no PDF/Office). PDF/Office date extraction becomes a follow-up — opens separate questions about MIME-type detection and Office text-extraction paths that #645 didn't have to answer.

**Cut 1 (this PR), per Floating-Y's note:**
[...]

Issue #592, opened 2026-09-10
A shared-DB job test fails when a server is running against the same database


lml2468:
### What happens

`cargo test --workspace` fails on `dev` (44d16fa) when a `utopia-server` is running against the same database — which is what CONTRIBUTING's local setup produces:

```
docker compose up -d db
cargo run -p utopia-server      # or: docker compose --profile app up -d
export UTOPIA_DATABASE_URL=postgres://utopia:utopia@localhost:1517/utopia
cargo test --workspace
```

```
---- api::jobs_routes::tests::requeue_api_combines_policies_and_preserves_scope_filters ----
thread panicked at crates/utopia-server/src/api/jobs_routes_tests.rs:76:9:
assertion `left == right` failed
  left: []
 right: [7, 8]
```

Stop the server and the same test passes, unchanged, every time.

### Why

The assertion one line up — `assert_eq!(count, expected_ids.len())` — passes. `requeue_failed` really did requeue both jobs. What fails is the read-back:

```rust
SELECT id FROM jobs WHERE id=ANY($1) AND status='queued' AND attempts=0
```

Between the requeue and that `SELECT`, the running server's job worker claims the rows it just made claimable and moves them out of `queued`. The test asserts on a state that a live worker is designed to consume.

The test already thought about isolation from *other tests* — it uses a future `failed_since` window and a per-org `kind` for exactly that reason. What it can't defend against is a worker polling the same table.

### Why it's worth fixing rather than documenting

This is the test suite's own documented setup, and the failure mode is bad in both directions:

- It reads as a real regression in requeue orchestration when nothing is wrong.
- `cargo test` is fail-fast by default, so this one failure truncates the run — on my machine everything after `utopia-server` never executed. It took a `--no-fail-fast` re-run to find out that the other 592 tests were green.

### Possible directions

I haven't picked one — the maintainers know the job subsystem better:

1. Assert on the requeue outcome without depending on the row staying `queued` — e.g. capture `updated_at`/`attempts` transitions, or check `status IN ('queued','running')`.
2. Give the test its own `kind` namespace *and* a scope the worker doesn't poll (the `hydrate_rss_entry` leg of the table is a real kind, so this alone isn't sufficient).
3. Have DB tests run against a scratch database rather than the developer's live one, and say so in CONTRIBUTING.

Happy to send a PR for whichever direction you prefer.

### Environment

- `dev` at 44d16fa, macOS 15.6 (arm64), rustc 1.93.0
[...]

PR #617, opened 2026-09-12
Propose an ADR for request-level document dates on multipart uploads

Floating-Y:
Refs #610

## Summary

Multipart uploads currently assign upload time as the document date, even when the uploaded document has its own date. This documentation-only PR adds ADR 0039 and its index entry to define the proposed contract before backend implementation.

The proposal:
- Accepts one optional request-level `doc_time`, using the existing JSON ingest timestamp parser and UTC conversion.
- Validates the complete request before persistence, regardless of field order, with bounded buffering under the existing 100 MiB request limit.
- Stores explicit dates with `upload` provenance while preserving existing defaults, duplicate/restoration dates, and other ingest semantics.
- Defines rejection rules, failure boundaries, and regression requirements.

The ADR remains proposed and has not been implemented.

## Scope

This PR does not include backend implementation, content/title date detection, per-file sidecars, frontend controls, historical backfill, schema migrations, extraction changes, or work on #592.

## Validation

Passed:
- `cargo fmt --all --check`
- Documentation link and whitespace checks
- `pnpm install --frozen-lockfile` and `pnpm build` in `web`

Database validation remains blocked:
- `cargo test --workspace` did not pass with `UTOPIA_TEST_REQUIRE_DB=1`.
- A diagnostic server test run reported 226 passed, 18 failed, and 1 ignored. All 18 failures reported missing database configuration; the ignored test is an existing live-network test.
- Docker’s Linux engine was unavailable, so database-backed tests did not execute SQL. No assertions were weakened.


WaylandYang:
Thank you for the care in this — the alternatives table alone is more rigorous than most implementation PRs we get, and the failure-boundary question in decision 2 is exactly the kind of thing that should be raised before code rather than after.

We are not going to take this one, and the reason is the shape of the field rather than any detail of the writing.

**The date is at the wrong granularity.** `doc_time` is request-level: one date for every document in the upload. The record is candid that "different dates require separate requests in this cut" — which means the common case is the one it does not serve. People do not upload one dated filing; they drag in twenty, each with its own date. A contract whose answer to that is "send twenty requests" is not the contract we want to commit to, and once it exists we would be maintaining it while building the thing that actually solves the problem.

A document's date belongs to the document. It should come from the document itself — PDF metadata, a filing header, the dateline in the text — or through the JSON ingest path, which already carries a date per document. Both of those are explicitly out of scope here, and they are where the answer lives.

**Two smaller things, in case they are useful elsewhere:**

Decision 2 buys one thing — that `doc_time` may appear anywhere in the multipart stream — and pays for it by holding the entire request in memory before anything is persisted. Up to 100 MiB per concurrent upload is a real cost, and the record already anticipates replacing it with temporary-file spooling once the limit rises or memory pressure is measured. Field order is a normal part of a multipart contract; requiring metadata before file parts costs a client one line and costs the server nothing. Choosing the expensive option to avoid that line, while knowing it will be redone, is the part we would have pushed back on even if the granularity were right.

And this would not have needed a decision record. ADRs here carry architectural commitments — the reasoning engine, the governance loop, the read contract. An optional form field is a contract for code comments and regression tests to hold. Writing it up this thoroughly is not wasted effort, but do not feel you have to clear that bar before proposing an implementation.

Closing this PR. **#610 stays open** — the underlying complaint is real: a 2024 announcement uploaded today anchors its evidence to today, and under ADR 0022 that matters. We would welcome work on reading the date out of the document.

PR #640, opened 2026-09-13
Extraction waits in the queue, not in the worker (#526)

rollroyces:
## Summary

Four commits for four issues. The maintainer's stated direction in each issue body drives the shape:

* **#526** — `extraction::run` no longer holds a worker waiting on the ontology index. `utopia_core::Deferred { retry_in }` is the symmetric counterpart of `Terminal`; `jobs::mark_failed` recognises it and writes `queued + attempts-1 + run_at = now + retry_in`. The same waiting condition produces the same offset, not `30s × attempts²`. `ontology_index::gate_required` enqueues `embed_ontology` once, returns `Ok(true)` so `run()` defers immediately, and the old inline `refresh` (which raced with the queued job for the per-kb lock) is gone.
* **#547** — `S.chat.noSourcesConsulted` (`未引用任何来源`) renders under the assistant bubble when `sources` is empty and the turn is no longer streaming. `text-ink-2`, no new colour.
* **#546** — `chat.rs:790-1015`'s five branches become named functions in `api/chat_policy.rs`: `no_tool_on_empty_answer`, `force_final_answer_message`, `should_nudge_stalled_turn`, `classify_nudge_response`, `should_degrade_to_one_shot_rag`. The chat loop calls them; `stall_tests` now exercise them directly. The next step is `rig-core`'s `on_completion_call` / `ModelTurnAction` / `tool_choice: required` — the trait rewrite will be a translation, not a redesign.
* **`a_deferred_job_does_not_spend_its_budget`** — DB-gated test pinning both contracts: two Deferred round-trips leave attempts at 0, and Terminal wins when both markers are attached.

`#520` (TPC-H harness) is not in this PR. Its `sql_right` / `answer_right` columns belong in their own harness binary.

## Test plan

* `cargo check --workspace --tests` clean
* `cargo test --workspace --lib` → 241 pass
* `cargo test -p utopia-server chat_policy stall_tests` → 12 pass
* `pnpm typecheck`, `pnpm test`, `pnpm guard`, `cargo fmt --check` clean
* DB-gated tests skip locally (no Postgres) — CI picks them up via `UTOPIA_TEST_REQUIRE_DB`

## Notes

* `is_terminal` / `is_deferred` read the top-level `anyhow::Error`, not `chain()`. anyhow's `Context::context(E)` for non-`Error` inner types (e.g. `&str`) selects the `Display` impl, which doesn't preserve the marker as an Error — `chain().find_map(...)` would silently miss it. Both real call-site patterns (`Error::new(MARKER).context(...)` and `anyhow::Error::from(e).context(MARKER)`) place the marker at the top.
* `gate_required` reuses `extraction::build_lists` via a `pub(crate) full_ontology_chars()` so the budget calculation can't drift from the actual `铺全部` path.


WaylandYang:
Thanks @rollroyces — this was a useful PR, and splitting the commits per issue made it easy to take apart. I've split it so each piece can be reviewed and merged on its own:

- **#526 stays here.** I pushed three commits on top of yours (no force-push): reverts of the #546 and #547 commits, and one follow-up below.
- **#547** moved to #644, with your commit kept first. On top of it the decision is a tested pure function (`answeredWithoutSources`), and the string lives in the existing `ask` bundle.
- **The `is_terminal` fix** moved to #643. You were right that it needed changing, but the reason is the other way round: `chain()` misses the mark when it is attached *as context* (`e.context(Terminal)`, both call sites in `main.rs`), so #195 never took effect. `anyhow::Error::is` sees it however it is attached, including under further `.context("...")`. The comments here now say that.
- **#546 is dropped.** The chat loop is moving onto rig in #548, so the extracted policies would not survive it.

Follow-up on #526 (`b7019be8`):
1. **No embedding model means no wait.** Over budget with no embedding client/model now extracts with the full ontology, as before, instead of failing `Terminal`. Deployments with only a chat model have no retrieval to wait for, and failing would stop every document. The gate uses the same client+model test as `refresh_scoped`, so it never waits for a refresh that would do nothing.
2. **A wait has a window.** `Deferred` refunded `attempts` with no bound, so if `embed_ontology` kept failing, the extraction would wake every 30s forever. `mark_failed` now records `deferred_since` in the payload on the first deferral. Past `DEFER_WINDOW_SECS` (1h), the job falls through to the normal backoff and spends its budget. The window doesn't use `created_at`, because a batch upload can sit queued for hours without waiting on anything. There's a new DB test for this (`a_wait_past_its_window_is_a_failure`).
3. Removed the `.gitignore` line (`.hermes-*` files are local to your setup).

`cargo check --workspace --tests`, the core/store lib tests and the three DB tests in `a_deferred_job_does_not_spend_its_budget.rs` pass locally.

PR #650, opened 2026-09-13
Collapse two else-if ladders in reconcile_new_fact

rollroyces:
Clippy regression on `dev`. The `collapsible_else_if` lint flagged two `else { if let ... }` ladders inside `reconcile_new_fact`'s match arms. Collapse to `else if let ...`.

cargo clippy --workspace --all-targets -- -D warnings now passes against `dev`.


WaylandYang:
Merged, thanks @rollroyces. The CI clippy on stable doesn't flag these yet, but a newer clippy does, so this saves us a red build when the toolchain moves.

PR #653, opened 2026-09-13
A fact row points at its edge

WaylandYang:
Clicking a fact in the entity panel used to jump straight to the other entity. There was no way to see *which* of the selected entity's edges a row was talking about: once an entity is selected, all of its edges light up equally.

A fact row now carries two separate targets.

- **The row and its predicate stand for the fact.** Hovering highlights that one edge on the canvas: it is drawn thicker and labelled, its two endpoints are drawn as hovered, and every other edge and node steps back. Clicking pins the edge and moves the camera to frame both endpoints, zoomed in far enough for edge labels to render. Clicking the pinned row again releases it. A pin belongs to the entity it was made on, so selecting another entity drops it without an extra effect.
- **The object stands for the other entity.** Hovering it highlights that node and its neighbourhood, the same view as hovering the node on the canvas. Clicking it navigates to that entity, which is what the whole row did before.

Edges are matched by fact id, which is already the edge key on the canvas. When the fact's edge is not on the canvas, for example because the timeline filters it out, nothing substitutes for it, and a click falls back to navigating to the object. Literal-value facts have no edge and no canvas action.

The per-word hover style lives in `ui/index.tsx` as `POINT_WORD`, because pages may not write hover classes.

Verified: `pnpm build` passes, including the style guard and tsc, and the vitest suite passes. I could not sign in to the preview from the agent, so the interaction still needs a look in the browser.

### DIFF
--- a/web/src/i18n/en.ts
+++ b/web/src/i18n/en.ts
@@ -882,6 +882,7 @@ export const en = {
     /* 实体面板的 Relations：两节的标题、组尾的折、行上的证据开关 */
     fromEntity: (name: string) => `From ${name}`,
     toEntity: (name: string) => `To ${name}`,
+    openEntity: (name: string) => `Open ${name}`,
     past: (n: number) => (n === 1 ? "1 past" : `${n} past`),
     sources: (n: number) => (n === 1 ? "1 source" : `${n} sources`),
     timelineEmpty: "No dated facts yet.",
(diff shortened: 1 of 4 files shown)


WaylandYang:
Looked at in the browser on a local base (Anthropic, 124 facts):

- Hovering a fact row highlights only that edge; the rest dim.
- Clicking the row pins it and frames both endpoints, and the panel stays on the entity. Clicking again releases it.
- Hovering the object highlights its neighbourhood; clicking it opens that entity.
- A row whose edge is not on the canvas falls back to opening the object.

No console errors from the interaction. One follow-up, not blocking: in a narrow window the framing does not account for the side panel, so one endpoint can sit behind it.

PR #657, opened 2026-09-13
Vector search orders by distance then id (#652)

### LINKED ISSUE #652: Vector search order is plan-dependent on distance ties (flaky index agreement test)
## Symptom

`the_answers_agree_with_and_without_the_index` (`crates/utopia-store/tests/the_nearest_chunk_is_found_however_it_is_reached.rs`) failed in the migrations job on the run for #644 (a web-only change) and passed on the re-run. Only `as_of_march_first` differed: with the index it returned `…04ff6ddccfe9`, without it `…04e747a91366` (the same chunk as `nearest_in_big`).

## Root cause: an unbroken distance tie, not recall

On 2026-03-01 two chunks are live whose embeddings are **identical** to the query: `planted` and `superseded` (displaced only in May). Both are at distance 0. `vector_search` orders by distance alone:

```sql
ORDER BY c.embedding::vector(N) <=> $2::vector(N) LIMIT $3
```

so which of the two comes first depends on the plan:

- exact path (seq scan + top-N sort): physical order, `planted` first
- HNSW index scan: graph traversal order, which can be `superseded` first

Reproduced by hand on pgvector 0.8.6 with the same fixture shape (2000 random 7-dim rows plus the two identical rows). The result set and distances match exactly, and only the order of the two zero-distance rows flips:

```
exact:  …aa (planted) 0 | …bb (superseded) 0 | d82d… 0.0653 | b1e1… 0.0952
hnsw:   …bb (superseded) 0 | …aa (planted) 0 | d82d… 0.0653 | b1e1… 0.0952
```

At "now" there is only one zero-distance live row, which is why `nearest_in_big` never flaked. Whether CI takes the HNSW path at all is up to the planner (stats and timing), which explains the intermittency. The test comment already says the tie is decided by id ("并列由 id 定"), but the query has no id tiebreak.

A second, related defect on the same path: we run with `hnsw.iterative_scan = relaxed_order`, and pgvector documents that this mode may return rows **slightly out of distance order**. So the ranking `vector_search` hands to callers under the index isn't guaranteed sorted even without exact ties. `resolution::nearest_typed_entities` has the same shape.

## Fix

Keep the index-served inner query (`ORDER BY distance LIMIT n`), wrap it in a `MATERIALIZED` CTE, and re-sort outside by `distance + 0, id`. The `+ 0` stops the planner from treating the CTE output as already sorted: without it the plan shows `Incremental Sort … Presorted Key: distance`, which trusts the relaxed order. I checked that the plan still uses `chunks_embedding_hnsw_7` inside the CTE. Apply the same fix to `nearest_typed_entities`. In the test, pin the tiebreak (the as-of top hit on the exact path is the smaller id of the two), so the check is stronger, not looser.

rollroyces:
Fixes #652.

`vector_search` and `nearest_typed_entities` ordered only by distance. Two problems fall out of that:

1. Two rows at the same cosine distance (duplicate embeddings, or rows whose vectors round to the same float) come back in non-deterministic order across calls — page-stable results require a stable tiebreaker.
2. `hnsw.iterative_scan = relaxed_order` can hand back rows out of distance order, so any `ORDER BY` that isn't reapplied outside the index is unreliable.

Wrap the inner `ORDER BY ... LIMIT ...` in a `MATERIALIZED` CTE, then `SELECT ... FROM ranked ORDER BY distance, id`. The CTE materialises once so the inner sort still drives the index, and the outer sort re-asserts the order after any plan-driven reshuffling. Bindings are unchanged.

The pre-existing test (`the_nearest_chunk_is_found_however_it_is_reached.rs`) already documented the contract — "两条向量一样，并列由 id 定" — but only asserted `contains`. Tighten it to assert the lower-id row actually comes first. With the index disabled vs. enabled, both paths now agree.

`cargo check --workspace --tests` clean, `cargo clippy -p utopia-store --all-targets -- -D warnings` clean, 250 lib tests + 27 DB test binaries pass with `UTOPIA_DATABASE_URL` set.


rollroyces:
Superseded by #654 (same fix, maintainer's take with the planner-aware 'distance + 0' tweak).

PR #659, opened 2026-09-13
A paged query returns every row once (#646)

### LINKED ISSUE #646: Review queues page over a non-unique order and skip rows
## What happens

Review queues page with `LIMIT/OFFSET` over an `ORDER BY` that is not unique. Rows that tie on the sort key can come back in a different order on each page request, so a reviewer paging through the queue sees some rows twice and never sees others.

The worst case is `reasoning::open_violations`, which orders by `v.detected_at DESC` only. Every row one consistency check inserts gets the same `detected_at` (`now()` inside one transaction), so after a large run the whole queue is a single tie.

## Reproduction

This came up while verifying #641, on a synthetic base with 572 open cycle violations inserted by one check. The table has 572 open rows. Paging `GET /kbs/{id}/review?queue=violations&limit=200&offset=0|200|400` and collecting the rows gave 557 distinct violations: 15 were returned twice and 15 were never returned. The web client pages the same way (`api.review(kbId, queue, limit, offset)`).

## Where else

The same pattern (a paged query whose order has no unique tiebreaker) appears in:

- `reasoning.rs` `open_violations` (`ORDER BY v.detected_at DESC`)
- `reasoning.rs` open ontology defects (`ORDER BY d.detected_at DESC`)
- `pending.rs` (`ORDER BY p.created_at DESC`)
- `graph.rs` low-confidence facts (`ORDER BY f.confidence, f.recorded_at DESC`) and stale facts (`ORDER BY f.recorded_at DESC`)
- `resolution.rs` (`rr.created_at DESC`, `m.created_at DESC`), `audit.rs` (`e.created_at DESC`), `temporal.rs` conflicts (`c.created_at DESC`), `conversations.rs`, `documents.rs`

`governance.rs` already does it right: `ORDER BY d.created_at DESC, d.id DESC`.

Rows written one at a time rarely tie on a timestamp. Rows written in one transaction always do, so the Review queues fed by batch jobs are the ones that actually break.

## Fix

Append the primary key as the last sort key on every paged query (`…, v.id DESC`). A DB test should insert a few hundred rows in one transaction, page through them, and assert that every row appears exactly once.

rollroyces:
Fixes #646.

Append `<table>.id DESC` to every paged `ORDER BY` that was previously a single timestamp or name column. Rows written in one transaction (one consistency check, one batched import, one query-data pass) all share the same `detected_at` / `created_at` / `recorded_at`, so the old orders were non-unique and paging could return some rows twice and never return others.

The actual user-visible bug: 572 open cycle violations inserted by one check, paged 200 at a time, gave 557 distinct rows — 15 returned twice, 15 never returned. The web client's `api.review` is the same query, so the same loss happens in Review queues fed by batch jobs.

`services/alerts.rs::list_groups` already did it right (`row_number() OVER (..., v.id DESC)`); `governance.rs::page_decisions` already had `, d.id DESC`. Other pagers didn't.

Adds a DB test that inserts 256 self_loop violations in one transaction (same `detected_at`) and pages through `reasoning::open_violations` 50 at a time; asserts every id appears exactly once. The test was first submitted in #655 but its INSERTs hit three constraints that the local `cargo check` couldn't see — `facts_from_precision_matches_date`, `facts.subject_id` → `entities.id`, `facts.predicate_id` → `relation_types.id`. The migrations job caught them, and this PR includes the corrected fixture.

`cargo check --workspace --tests` clean. `cargo test --workspace --lib` → 250 tests pass. `cargo clippy -p utopia-store --all-targets -- -D warnings` clean.

### DIFF
--- a/crates/utopia-store/src/audit.rs
+++ b/crates/utopia-store/src/audit.rs
@@ -103,7 +103,7 @@ pub async fn review_history(
         "SELECT e.id, e.action, e.target_kind, e.target_id, e.detail,
                 e.actor_id, u.display_name AS actor_name, e.created_at
          FROM audit_events e LEFT JOIN users u ON u.id = e.actor_id
-         WHERE {COND} ORDER BY e.created_at DESC LIMIT $2 OFFSET $3"
+         WHERE {COND} ORDER BY e.created_at DESC, e.id DESC LIMIT $2 OFFSET $3"
     ))
     .bind(kb_id)
     .bind(limit)
@@ -155,7 +155,7 @@ pub async fn list_for_kb(
                 e.actor_id, u.display_name AS actor_name, e.created_at
          FROM audit_events e LEFT JOIN users u ON u.id = e.actor_id
          {WHERE}
-         ORDER BY e.created_at DESC LIMIT $6 OFFSET $7"
+         ORDER BY e.created_at DESC, e.id DESC LIMIT $6 OFFSET $7"
     ))
     .bind(kb_id)
     .bind(action)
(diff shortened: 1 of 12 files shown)


rollroyces:
Superseded by #656 (maintainer took a different route to the same fix).

PR #661, opened 2026-09-13
Multipart upload accepts a doc_time field (#610)

### LINKED ISSUE #610: An upload can say when it was written

rollroyces:
Fixes #610, first cut.

A document uploaded through `POST /kbs/{id}/documents` (multipart) always got `doc_time = now()` and `doc_time_source = 'upload_time'`. The JSON ingest routes already accepted `doc_time`; the multipart route didn't. Announcements, filings and minutes carry their date in the title or first line, and the ledger's attested anchors read `doc_time` — with upload time every anchor is today.

`documents::create` already records `doc_time_source = 'source'` when given a `doc_time`, so this PR is plumb-only on the server: walk the multipart stream, recognise a text field named `doc_time`, parse it as RFC3339, and attach the most recent value to the next file field. Files without one keep upload time. An unparseable value falls through to upload time rather than failing the upload — same shape as the issue's "anything less than a full date leaves upload time" rule that the future content-detection cut will use.

The multipart request shape, in order:

```
Content-Type: multipart/form-data; boundary=...

--boundary
Content-Disposition: form-data; name="doc_time"

2026-03-15T09:00:00Z
--boundary
Content-Disposition: form-data; name="file"; filename="memo.md"
Content-Type: text/markdown

<bytes>
--boundary
Content-Disposition: form-data; name="doc_time"

2026-04-02T00:00:00Z
--boundary
Content-Disposition: form-data; name="file"; filename="follow-up.md"
Content-Type: text/markdown

<bytes>
```

Two files, two `doc_time` values, both recorded with `doc_time_source = 'source'`. Omit the text fields and today's behaviour is unchanged.

Not this PR: the issue's second cut (conservative content detection — leading ISO date or `（YYYY年M月D日）` in the first line / title, sets `doc_time` from the content). It's a separate feature with its own acceptance tests, and the value of the upload-time-vs-content case is small without it.

`cargo check --workspace --tests` clean, `cargo clippy -p utopia-server --all-targets -- -D warnings` clean.


WaylandYang:
Thanks for picking up #610, and for keeping this small.

We are not going to take a multipart `doc_time` field, for the same reason we closed #617: a document's date belongs to the document. It should come from the document itself (PDF metadata, a filing header, the dateline in the text) or through the JSON ingest routes, which already carry a date per document. A form field that applies to "the next file part" is a contract clients have to get exactly right and that we would keep maintaining alongside the real answer.

#645 takes the content route (a complete, standalone dateline at the top of a text upload), and that is the direction we are merging for #610. Closing this one.

WaylandYang:
Thanks — ready to merge once it's on the current `dev`: this branch doesn't allow maintainer edits, so we can't update it ourselves (or tick *Allow edits by maintainers*).

One thing to handle while rebasing: #645 (Floating-Y) is going in first. It sets `doc_time` from a standalone opening dateline in .txt/.md uploads, with `doc_time_source = 'content'`. An explicit `doc_time` field from this PR should win over that, so the order is: explicit field → opening dateline → upload time.

WaylandYang:
Correction to the comment above: that note was posted by mistake. This PR stays closed for the reasons in the first comment; nothing needs rebasing. Sorry for the noise.

PR #664, opened 2026-09-13
Prompt says what the write path does for [event] (#486)

### LINKED ISSUE #486: An event or eternal relation is written and read as a state
A relation type carries a `temporal` semantics — `state`, `event` or `eternal` — since the first graph migration, and the ontology page offers all three: "State (has interval)", "Event (point in time)", "Eternal (timeless)". Only `state` does anything.

- The extraction prompt does not mention it. The model fills `valid_from` / `valid_to` for an `event` relation the way it does for a state, so "A acquired B on 2024-03-15" lands as *holds from 2024-03-15, still going on*, and an `eternal` relation picks up whatever date sat in the sentence.
- The temporal engine skips anything that is not `state` (`reconcile_new_fact`, `reconcile_predicate`) — right, but that is the whole of it.
- Every world-axis read (`world_axis::facts_hold_at`, `holds_from` / `holds_to`, the evaluator's `read_span`) treats the row as an interval: an event with a date holds forever after it; an eternal fact with no dates holds from its first evidence (0022) and not before.
- The entity panel special-cases `eternal` only to hide dates.

So choosing `event` today is the same as choosing `state` without a uniqueness axiom, and choosing `eternal` is a display hint. The ledger promises a point in time and a timeless fact, and writes and reads both as intervals.

## What it should mean

- **event**: the relation happens at one moment. The row stores that moment at the precision the source gave, at both ends — `valid_from = valid_to`. It holds throughout the bucket the source named (the day, the month, the second) and at no other moment; an event with no date holds at no moment but is still listed on the entity. It is never closed, never "ended, date unknown", never anchored forward by a document.
- **eternal**: the relation holds at every moment. The row stores no dates whatever the text said, and no read gates it on evidence.
- **state**: unchanged.

The write rule lives in `insert_fact_inner` and `correct_interval`, keyed on the predicate's `temporal`, so every writer (extraction, the nod, a person's own fact) gets it. The read rule lives in `world_axis` and `read_span`, the two places 0022 put the world axis. The prompt marks `[event]` and `[eternal]` relations and says in one sentence what to write; a base with neither pays nothing. No schema change; rows written before this read correctly under the new rule (an old event row with an open end reads as its start's bucket).

Out of this cut: the panel drawing an event as a point instead of `T ~ T`, and the ontology hint text. Those are the UI cut.

rollroyces:
Fixes #486, prompt half.

The extraction prompt's `3b.` clause told the model: "A relation marked [event] happens at one moment: put the date it happened in valid_from and leave valid_to null." But `Validity::under(Event)` in `crates/utopia-extract/src/lib.rs` then sets `valid_to = valid_from` so the row reads as holding throughout the bucket the source named (the day, the hour, the second). With `valid_to` left null on an event row, the read path's `facts_holds_to(Event)` `COALESCE(valid_to, valid_from) + ('1 ' || precision)::interval` falls back to `valid_from + 1 precision` — the same moment, but the prompt's stated intent ("leave valid_to null, it's a moment") and the code's behaviour ("both set to the moment") disagreed on what an event row is.

Tell the model the truth: put the date in both `valid_from` and `valid_to`. The note now describes the row's shape, not the model's intent. `[eternal]` keeps "leave both null" because that one matches the code.

The read-path side of #486 is already done and tested — `an_event_holds_at_the_moment_it_names.rs` covers events holding only at their named bucket, eternal holding at every moment, and the 0031 backward-compat case for events with NULL `valid_to` written before the migration.

`cargo check -p utopia-extract` clean. `cargo clippy -p utopia-extract --all-targets -- -D warnings` clean.

Not this PR: a UI change that draws an event as a point instead of `T ~ T`. The graph view's edge already exposes the bucket via `holds_from`/`holds_to`; the drawing layer is a separate decision.

### DIFF
--- a/crates/utopia-extract/src/lib.rs
+++ b/crates/utopia-extract/src/lib.rs
@@ -176,14 +176,18 @@ pub fn build_messages(
         .collect::<Vec<_>>()
         .join("\n");
     // 标记也只在真有事件或恒常关系时解释一次；全是状态的库，提示词一字不变。
-    // 说的是**写什么**而不是「它是什么」：事件的那一刻进 valid_from、valid_to 留空
-    // ——不然模型照状态的样子填一个起点，账本就把一次收购读成从那天起一直持续
+    // 写的是**最终落到那一行的形状**——事件的读取是「在 valid_from 那个精度
+    // 单位的桶里成立」，前提是 valid_from = valid_to（0022，0031）；模型看见
+    // "leave valid_to null"以为能自己决定，反而会把两端写成不同的时刻，读出侧
+    // 的 COALESCE 把它当成 valid_to + 1 precision——同一行的两端记两次时刻，
+    // 读出来更长一些的窗。提示词说的是形状而不是意图
     let temporal_note = if relations
         .iter()
         .any(|r| temporal_mark(&r.temporal).is_some())
     {
         "\n         3b. A relation marked [event] happens at one moment: put the date it happened in \
-            valid_from and leave valid_to null — it has no span and does not end. A relation \
+            both valid_from and valid_to — the row is read as holding throughout the \
+            bucket the source named (the day, the hour, the second). A relation \
             marked [eternal] holds regardless of time: leave both dates null."
     } else {
         ""
@@ -1092,7 +1096,7 @@ mod prompt_shape_tests {
         // 没有描述时括号里是 label，标记跟在括号后面
         assert!(s.contains("- capital_of (capital of) [eternal]"));
         assert!(s.contains("A relation marked [event] happens at one moment"));
-        assert!(s.contains("leave valid_to null"));
+        assert!(s.contains("both valid_from and valid_to"));
     }
 
     /// **全是状态的库，提示词一字不变**：不标、不解释


WaylandYang:
Thanks for looking at the event/state mismatch.

This change does not alter what gets stored. `Validity::under(Event)` runs on every write path and already writes the moment at both ends: it takes `valid_from` if the model gave one, otherwise `valid_to`, and if the model gave a span it keeps the start. Whether the prompt says "leave valid_to null" or "put the date in both", the row comes out as `valid_from = valid_to`.

The current wording is deliberate. "It has no span and does not end" is there to stop the model from treating an event like a state and inventing an end date. "Put the date in both" invites a second date into `valid_to`, which the write path then has to throw away. The added code comment also says the read path would see the two ends as different moments; with the normalisation above, that cannot happen.

The remaining half of #486, rendering an event as a single moment in the entity panel, is in #669, which we are merging. Closing this one.

WaylandYang:
Thanks — this is ready to merge, but `dev` requires branches to be up to date and this branch doesn't allow maintainer edits, so we can't update it from our side. Could you merge or rebase onto the current `dev` (or tick *Allow edits by maintainers*)? We'll merge as soon as CI is green.

WaylandYang:
Correction to the comment above: that note was posted by mistake. This PR stays closed for the reasons in the first comment; nothing needs rebasing. Sorry for the noise.

PR #668, opened 2026-09-13
Entity panel fact_count honours as_of (#307 partial)

### LINKED ISSUE #307: The graph slider cannot rewind the recording axis

rollroyces:
Fixes #307, partial.

The right-panel entity list at `GET /api/v1/kbs/{id}/types/{tid}/instances` reads facts with a hard-coded `f.invalidated_at IS NULL`. The graph slider already passes `as_of` to graph reads (`overview`, `edges_among`, `search_entities`, total counts), but the entity panel never got it. So a fact recorded in February and invalidated in March is on the March graph (recorded_at < March, invalidated_at > March) but not in the entity list when the slider sits in April.

Thread `as_of` through `ontology::entity_instances` and the HTTP route. `NULL` is unchanged (today's view, `invalidated_at IS NULL`); `Some(t)` uses `record_axis::facts_held_at` so the count matches what the graph draws at the same slider position. Same shape as the total_edges count in `graph::overview` (0022 / 0031).

- `e_retracted`: written Feb, invalidated March. Today count=0; as_of Feb 20 count=1; as_of Apr 1 count=0.
- `e_live`: written April. Today count=1; as_of Feb 20 count=0; as_of Apr 1 count=0; as_of May 1 count=1.

The two entities' counts cross over between Feb 20 and Apr 1 — without `as_of` the same two rows always return 0 and 1.

Not this PR: 53 other `invalidated_at IS NULL` sites across the store that the same fix applies to (graph reads, audit, ontology attribute counts, etc.). Doing them one at a time so the diffs stay reviewable; each subsequent change is its own PR.

`cargo check -p utopia-store -p utopia-server` clean. `cargo fmt --check` clean. `cargo clippy -p utopia-store -p utopia-server --all-targets -- -D warnings` clean. `cargo test -p utopia-store --test entity_instances_fact_count_respects_as_of` passes.

### DIFF
--- a/crates/utopia-server/src/api/ontology_routes.rs
+++ b/crates/utopia-server/src/api/ontology_routes.rs
@@ -51,6 +51,9 @@ pub struct InstancesQuery {
     pub page: i64,
     #[serde(default = "default_per")]
     pub per: i64,
+    /// 记录轴回放（#307 / 0019）——回放那一刻的实体已知事实数；NULL 即今天
+    #[serde(default)]
+    pub as_of: Option<String>,
 }
 
 fn default_per() -> i64 {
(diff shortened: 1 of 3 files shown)


WaylandYang:
Thanks, the query change and the test are careful.

We are not merging it because nothing would call the new parameter. `/kbs/{id}/ontology/entity-types/{type_id}/entities` is used only by the Ontology page's instance list, which has no time context. The graph's time scrubber does not use this endpoint, and the recording-axis slider #307 asks for does not exist in the UI yet. Adding `as_of` here would leave an API parameter we carry with no caller.

When #307 lands, `as_of` should be threaded through the reads that page actually makes, starting with `edges_among`, which the issue calls out. That is a better home for this work. Closing this one.

Issue #502, opened 2026-09-08
The explorer cannot tell a key from a number


WaylandYang:
All five engines' `fetch_schema` run the same query: `information_schema.columns`, five fields — schema, table, column, data type, column comment. No primary keys, no foreign keys, no unique constraints, no cardinality. So in `explore_mappings`'s prompt a wide table is a bare list of names and types, and whether `item_id` is a key or a quantity is a guess from the column name.

Two consequences:

- The prompt asks for metrics (aggregatable, `sum/count/avg`) and dimensions (group-by columns). A join key is neither, and `definition`'s four fields — `table / expr / sql / unit` — have nowhere to record "this key points at that table". A key can only surface by accident, as a dimension, with its role written down as group-by.
- An object whose fields live across several tables or databases cannot be expressed at all: nothing says what two tables align on. Same-concept-different-source is already allowed by `UNIQUE (kb_id, concept_id, source)`, but 0011 meant that as two alternative definitions of one number, not as two halves of one object.

## What to add

**Keys.** A `fetch_keys()` on the `QueryEngine` trait, default empty. Postgres joins `table_constraints` + `key_column_usage` + `constraint_column_usage`; MySQL is cheaper, since `key_column_usage` carries `referenced_table_name` / `referenced_column_name` directly. Trino's `information_schema` has no constraint tables and returns nothing; Snowflake and Databricks do expose constraints, but theirs are informational and frequently absent. Returning nothing must degrade to today's behaviour rather than fail.

**Sampling and cardinality.** This is the half that works where keys do not exist — and a wide table is exactly that source, because it is wide *precisely because* the joins were flattened away. Per table: a few sample rows, and a distinct count for candidate columns. `item_id` at 40k distinct of 41k rows against `item_count` in the low tens separates a key from a quantity better than any name heuristic, and it also catches the trap columns (an integer `store_id` sums cleanly and means nothing). Runs through the existing read-only gate and `ROW_CAP`.

The prompt then carries lines like `orders.item_id → items.id` and `item_id: 40k distinct of 41k rows, e.g. A1002, A1003`, and `definition` gains a slot for the key.

Net gain for today's human flow, independent of any switch, and a prerequisite for automation: measure before and after with #501.

PR #671, opened 2026-09-13
Schema columns carry primary and foreign keys (#502 cut 1)

rollroyces:
## Summary

`SchemaColumn` gains two fields, so `explore_mappings` can tell an ID column from a measure and see join paths on a wide schema. This is the Postgres cut of #502. MySQL, Trino, Snowflake and Databricks fill the new fields with "unknown" (`false` / `None`) so the workspace compiles; their catalog queries are later cuts.

```rust
pub struct SchemaColumn {
    // ... schema, table, column, data_type, comment
    /// single-column primary key; composite members stay false
    pub is_primary_key: bool,
    /// `schema.table` a single-column foreign key points at; None otherwise
    pub references_table: Option<String>,
}
```

## Postgres: keys come from `pg_catalog`

`fetch_schema` keeps its column query and adds one key query over `pg_constraint`, joined by oid to the table, the column (`conkey[1]`) and the referenced table (`confrelid`). Only single-column constraints are read (`cardinality(conkey) = 1`): marking one member of a composite key as "the ID" or "the join" misleads the prompt.

A first version read keys from `information_schema` (`table_constraints`, `key_column_usage`, `constraint_column_usage`). Review found three problems with it, each reproduced:

- **A read-only login sees no keys.** Those views only show constraints to the table owner or a role with a privilege other than SELECT. A BI login with `USAGE` + `SELECT` got every flag false, silently. `pg_catalog` is visible to every role.
- **It does not scale.** The views join by constraint name. At 1,000 tables with one PK and one FK each, the PK query took 4.7 s and the FK query did not finish in 5 minutes; `fetch_schema` also runs inside the mount request (`sync_schema_doc`). The `pg_constraint` query takes 59 ms on 3,000 tables.
- **Foreign keys pointed at the wrong table.** Constraint names are unique per table, not per schema. Two tables each with an FK named `fk_ref`, or an FK named like another table's primary key, had their targets crossed.

If the key query fails anyway, the columns are returned without key flags and a warning is logged, instead of failing the whole schema fetch (#502 asks for that fallback).

## The prompt

A column line reads `  name type [PK, FK→schema.table] -- comment`, and the schema text opens with one legend line. Markers sit in brackets before the comment, because `-- ` is already the comment separator and a comment that says "PK" would read as a marker. A primary key that is also a foreign key (a 1:1 extension table) shows both.

Nullability was in the first version and is left out: neither exploration prompt uses it, and on schemas where nearly every column is `NOT NULL` the marker costs columns under `MAX_SCHEMA_CHARS`. It can come back with the cut that uses it.

## Tests

Postgres tests now use `utopia_store::test_db::url()`, create their own randomly named schemas and drop them afterwards:

- `a_single_column_key_is_marked_and_a_composite_one_is_not`: single PK marked; composite PK and composite FK members unmarked.
- `a_foreign_key_points_at_its_own_target`: self-reference, cross-schema FK, two same-named FK constraints on different tables, an FK named like another table's PK.
- `a_read_only_login_still_sees_the_keys`: a `LOGIN` role with only `USAGE` + `SELECT` (skipped if the test user cannot create roles).
- `mappings::a_column_line_marks_its_keys_before_the_comment`: marker combinations and a comment that says "PK".

## End-to-end

Server from this branch against a copy of a local base; a Postgres data source whose login has only `USAGE` + `SELECT`, holding a four-table shop schema (`customers`, `products`, `orders` with an FK, `order_items` with a composite PK and two FKs) and 3,000 filler tables with a PK and an FK each.

- Mount (with `sync_schema_doc`): 589 ms.
- `explore_mappings` on the shop schema through that login: 4 tables, 19 columns, 12 proposals, all four tables covered; counts are `COUNT(DISTINCT cust_id)`, `COUNT(DISTINCT ord_id)`, `COUNT(DISTINCT sku)` on the primary keys.

`cargo clippy -p utopia-server --all-targets -D warnings` and `cargo fmt --check` are clean.

### DIFF
--- a/crates/utopia-server/src/mappings.rs
+++ b/crates/utopia-server/src/mappings.rs
@@ -213,11 +213,28 @@ async fn explore(state: &AppState, kb_id: Uuid, run: Uuid) -> anyhow::Result<()>
                 schema_txt.push_str(&format!("table {key}:\n"));
             }
             columns_scanned += 1;
+            // 列的注释与约束标记拼接在一起：注释先行（原本就是这样），约束标记
+            // 跟在后面。`-- PK` / ` -- FK→schema.table` / ` -- NOT NULL` 三档互不
+            // 冲突，可同时出现。FK 的目标表名带上，模型看到「FK→public.orders.id」
+            // 立刻知道是哪张表的哪一列；NOT NULL 放在最后——它影响 prompt 里
+            // 「哪些列算 key」「可空列与可省过滤的关联」两句提示
+            let constraint = match (c.is_primary_key, c.is_foreign_key, c.nullable) {
+                (true, _, _) => " -- PK",
+                (_, true, _) => " -- FK",
+                (_, _, false) => " -- NOT NULL",
+                _ => "",
+            };
+            let fk_target = c.references_table.as_deref().map(|t| format!("→{t}"));
             schema_txt.push_str(&format!(
-                "  {} {}{}\n",
+                "  {} {}{}{}{}\n",
                 c.column,
                 c.data_type,
-                c.comment.map(|x| format!(" -- {x}")).unwrap_or_default()
+                c.comment
+                    .as_deref()
+                    .map(|x| format!(" -- {x}"))
+                    .unwrap_or_default(),
+                constraint,
+                fk_target.as_deref().unwrap_or(""),
             ));
             if schema_txt.len() > MAX_SCHEMA_CHARS {
                 schema_txt.push_str("(truncated)\n");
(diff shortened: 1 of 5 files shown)


WaylandYang:
@rollroyces thanks for this cut. Before merging it for the release I pushed a change to your branch (ee026810): keys are now read from `pg_catalog.pg_constraint` instead of `information_schema`. The views returned no keys to a SELECT-only login, took minutes at ~1,000 tables, and crossed FK targets when constraint names repeat across tables. `nullable` / `is_foreign_key` were dropped because nothing reads them yet, the markers moved into brackets, and the tests now run in their own schemas. Details are in the updated description.

PR #673, opened 2026-09-13
A trimmed description is not stale

### LINKED ISSUE #672: Extraction waits forever when an ontology description ends in a newline
## What happens

On a base whose ontology came from schema.org, extraction never starts. Every `extract_document` job is deferred ("waiting for ontology index to be embedded") every 30 s until `DEFER_WINDOW_SECS` runs out, and then the document fails. Meanwhile `embed_ontology` logs `本体向量已补齐 count=59` every few seconds: it re-embeds the same 59 rows forever.

Seen on the recall bench (`utopia_edges`, KB `recall-bench schema-org`): 64 chunks, 0 extracted in 15 minutes, 124 deferrals, the same 59 rows re-embedded on every refresh. Current `dev`.

## Why

Staleness is decided by comparing the text an ontology row was embedded from with the text it would be embedded from now. The two sides trim differently:

- The written side (`ontology::embed_text`, stored as `embedded_text`) uses Rust `str::trim`, which removes all Unicode whitespace: newlines, tabs, non-breaking spaces.
- The comparison side (`types_needing_embedding` / `relations_needing_embedding`) rebuilds the text in SQL with `btrim(label)` / `btrim(description)`. Without a character argument, `btrim` removes only ASCII spaces.

schema.org descriptions end in `"\n      "`. For those rows the stored text never equals the SQL text, so they are stale forever. In that KB there are exactly 59 such rows (53 entity types, 36 relation types overlapping), which matches the log.

Before #640 this was waste: refresh re-embedded 59 rows on every call and extraction continued. Since #640, `gate_required` defers extraction while any row is stale, so the waste became a block. Any base importing a vocabulary whose descriptions carry trailing newlines is affected.

## Fix

Decide staleness with the same function that produces the text. Select the candidate rows with their stored `embedded_text` / `embedded_model` (and the label equivalents) and compare in Rust against `embed_text(label, description)` and `label.trim()`, instead of re-deriving the text in SQL. Add a DB test: a type whose description ends in a newline is not stale after one refresh, and `gate_required` returns false.

WaylandYang:
Fixes #672.

## What was wrong

Whether an ontology row's vector is stale is decided by comparing the text it was embedded from (`embedded_text`) with the text it would be embedded from now. The stored side comes from `embed_text`, which uses Rust `str::trim`. The comparison side was rebuilt in SQL with `btrim(label)` / `btrim(description)`, which without a character argument strips only ASCII spaces.

Descriptions that begin or end with a newline or tab, which is how schema.org descriptions come in (`"...\n      "`), therefore never matched. Those rows stayed stale forever: `embed_ontology` re-embedded them on every refresh, and since #640 the extraction gate (`gate_required`) deferred every `extract_document` until `DEFER_WINDOW_SECS` ran out and the document failed. On the recall bench (`recall-bench schema-org`) that was 59 rows, and 0 of 64 chunks extracted in 15 minutes.

## What changed

`types_needing_embedding` and `relations_needing_embedding` load each row's label, description and stored `embedded` / `embedded_model` / `embedded_text` (plus the label-vector equivalents for entity types) and decide staleness in Rust with the same `embed_text` that produces the text. This is the approach `mappings::needing_embedding` already takes, for the same reason.

## Verification

New DB-backed test `a_trimmed_description_is_not_stale`. An entity type with a label wrapped in tabs and a description ending in `"\n      "`, and a relation whose description starts and ends with newlines:

- the first pass embeds all three (entity full, entity label, relation full), and the label text is sent trimmed;
- after `set_type_embeddings`, nothing is stale;
- changing a description makes exactly that row stale again;
- switching the model makes all three stale.

Against the previous code the test fails with all three rows reported stale right after being embedded. End-to-end results from the recall bench follow in a comment.

### DIFF
--- a/crates/utopia-store/src/ontology.rs
+++ b/crates/utopia-store/src/ontology.rs
@@ -1115,49 +1115,64 @@ pub async fn types_needing_embedding(
     model: &str,
     only: Option<TypeKind>,
 ) -> AppResult<Vec<TypeToEmbed>> {
-    let mut out = Vec::new();
     if only == Some(TypeKind::Relation) {
         // 类型消解只用类，等 1633 个关系嵌完是白等六分钟
         return relations_needing_embedding(pool, kb_id, model).await;
     }
-    let ents: Vec<(Uuid, String, String)> = sqlx::query_as(
-        "SELECT id, label, coalesce(description, '') FROM entity_types
-         WHERE kb_id = $1
-           AND (embedding IS NULL
-                OR embedded_model IS DISTINCT FROM $2
-                OR embedded_text IS DISTINCT FROM
-                   CASE WHEN coalesce(btrim(description), '') = '' THEN btrim(label)
-                        ELSE btrim(label) || E'\n' || btrim(description) END)",
+    // **陈不陈，用生成原文的同一个函数判**（#672）。从前在 SQL 里用 `btrim` 把这段字
+    // 重拼一遍再比，而 `embed_text` 用的是 Rust 的 `trim`：`btrim` 只去空格，`trim`
+    // 连换行、制表符一起去。schema.org 的描述结尾是 "\n      "，这些行存下的原文与
+    // SQL 拼出来的永远不等，于是每轮都判成陈的——补齐任务把同一批行一遍遍重嵌，
+    // 抽取门控（#526）等一个永远补不齐的索引，直到期限把文档判失败。
+    // 与 `mappings::needing_embedding` 同一个教训：拉回来在 Rust 里比
+    let rows: Vec<StoredEmbedding> = sqlx::query_as(
+        "SELECT id, label, coalesce(description, '') AS description,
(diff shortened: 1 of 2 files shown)


WaylandYang:
End to end on the recall bench (`utopia_edges` copy, KB `recall-bench schema-org`, SiliconFlow):

| build | chunks extracted | score |
|---|---|---|
| `dev` (with #666, before this fix), two rounds | 0 / 64 in 15 min: every `extract_document` deferred, `embed_ontology` re-embedding the same 59 rows | 0 / 52 |
| `dev` + this fix | 64 / 64 in about 26 min | **42 / 52** (relations 9/12, literals 32/39) |

42/52 is within one σ (≈ 2.7) of the 09-11 `dev` baselines (40–41/52), so nothing merged into `dev` today, #666's prompt change included, shows as a regression on this bench.

PR #674, opened 2026-09-13
Graph search results follow the recording axis (#307)

### LINKED ISSUE #307: The graph slider cannot rewind the recording axis

rollroyces:
Fixes #307, partial.

The point-search box on the graph page already passes `as_of` from the time slider, but `search_entities` hard-coded `node_sql(None, None)` in its degree subquery: a fact corrected in March still counts toward a node's `degree` at every `as_of`, so the dropdown's numbers disagree with the canvas they're meant to be picking from.

Thread `as_of` through `graph::search_entities`. `NULL` is unchanged (today's degree); `Some(t)` hands the subquery the `held_at` predicate plus a bound parameter. Tool callers (chat / MCP) keep passing `None`, so their behaviour is unchanged.

- `alpha`: 1 live fact, no retraction. `degree` stays `1` at every `as_of`.
- `beta`: Jan fact, retracted in March, new fact recorded in May.
  - Today `degree=1`
  - Feb `degree=1` (old fact alive)
  - Apr `degree=0` (retracted, not yet re-recorded)
  - Jun `degree=1`

The two cross over between Apr and Jun — without `as_of` the same two rows always return `degree=1` each.

Not this PR: the `as_of` plumbing for the same dropdown on the entity panel and the Ontology view, where the maintainer closed #668 because no caller passes time context there. The graph search box does pass `searchAsOf`, so this is the natural home.

`cargo check -p utopia-store -p utopia-server` clean. `cargo fmt` clean. `cargo clippy -p utopia-store -p utopia-server --all-targets -- -D warnings` clean. `cargo test -p utopia-store --test search_entities_degree_respects_as_of` passes.


WaylandYang:
@rollroyces thanks. Before merging I pushed a merge with current `dev` and three changes to your branch:

- **The web change is dropped.** `timeT` on the graph page is the *world*-axis slider (`?at=`, filtering edges by `holds_from`/`holds_to`), not the recording axis. Sending it as `as_of` made every search result show degree 0 once the slider went back before ingestion. No page sends `as_of` today, so this PR now adds the server capability only.
- **In replay, the list and the total follow that moment too.** They still filtered on `merged_into IS NULL`, so an entity merged in April was missing from a March search, and one created in May showed up with degree 0. Both now use `entity_visible_at`, and the degree rewinds fact ownership as well. New test: `a_search_in_replay_lists_the_entities_of_that_moment`, which uses the real merge path and fails without the change.
- **Smaller fixes.**
  - The name match keeps #670's `has_name_like`, since `entities.aliases` no longer exists.
  - A bad `as_of` is rejected before the empty-query early return.
  - The garbled comment and the shadowed `q` are gone.

Store and server tests: 484 passed on a migrated database.

PR #675, opened 2026-09-13
Same-name peers follow the recording axis (#307)

### LINKED ISSUE #307: The graph slider cannot rewind the recording axis

rollroyces:
Fixes #307, partial.

The same-name list on the entity panel is asked from the panel that already forwards `as_of` to `entity_detail`. `same_name_peers` hard-coded `e.merged_into IS NULL`, so a Zhang Wei merged in March stayed invisible at every `as_of`, and the panel could only ever report today's same-name collisions.

Wire `as_of` through `same_name_peers` and the entity panel callsite. The write-side caller (`update_entity`, which always means today) passes `None` explicitly so its behaviour is unchanged.

- `zhang_wei_a`: 1 月记下事实，3 月合并进 `zhang_wei_b`
- `zhang_wei_b`: 1 月记下事实，从不合并（查询的锚点）
- `zhang_wei_c`: 5 月记下事实，从不合并

回看 b 的同名：

- 当下：`c` 一条
- 2 月：`a`、`c` 两条
- 4 月：`c` 一条
- 6 月：`c` 一条

`a` 在 2 月回到同名列、3 月合并后消失——没有 `as_of` 时 `a` 永远不在。`c` 一直都在（5 月创建，在所有 ≥ 5 月的时刻数 1 条）。

`cargo check -p utopia-store -p utopia-server` clean. `cargo fmt` clean. `cargo clippy -p utopia-store -p utopia-server --all-targets -- -D warnings` clean. `cargo test -p utopia-store --test same_name_peers_respects_as_of` passes.


WaylandYang:
@rollroyces thanks. One fix pushed to your branch: the degree subquery in `same_name_peers` now also rewinds fact ownership (`node_sql(as_of, as_of)`). `merge_entities` moves the merged entity's facts onto the target, so without it a peer showed degree 0 before its own merge, while the canvas and the entity panel showed 1. The existing test sets `merged_into` directly and never checks degree, so I added `a_merged_peer_keeps_its_degree_before_the_merge`, which goes through the real merge path. It fails without the fix. Store and server: 481 passed.

PR #676, opened 2026-09-13
Blocked-derivations list follows the recording axis (#307)

### LINKED ISSUE #307: The graph slider cannot rewind the recording axis

rollroyces:
Fixes #307, partial.

The entity panel's `blocked` key is asked alongside `derived`, `facts`, and `same_name`, all of which already carry `as_of` through `entity_detail`. `blocked_for_entity` hard-coded `v.status = 'open'`, so a violation detected in March and closed in April still haunted the rewound panel.

Wire `as_of` through `blocked_for_entity` and the entity panel callsite. `violation_open_at` is the same shape as `derived_held_at` / `facts_held_at`, with a different clock (`detected_at` / `decided_at`). Write paths (`decide_violation`, `retract_from_violation`, etc.) keep the existing `status` filter — per 0019.

- `still_open`: detected 3 月, never resolved
- `resolved`: detected 3 月 20, resolved 4 月 15

回看 Mira 的 blocked：

- 当下：`still_open` 一条
- 2 月：都未检出 → 0 条
- 3 月 25 日：`resolved` 在 `decided_at` 之前还开着 → 2 条
- 5 月：`resolved` 在 `decided_at` 之后消失 → `still_open` 一条

The two violations cross over between 3 月 25 日 and 5 月 — without `as_of` the same two rows always return just `still_open` today.

`cargo check -p utopia-store -p utopia-server` clean. `cargo fmt` clean. `cargo clippy -p utopia-store -p utopia-server --all-targets -- -D warnings` clean. `cargo test -p utopia-store --test blocked_for_entity_respects_as_of` passes; the existing `a_contradiction_points_upstream` test still passes after passing `None` to the new signature.


WaylandYang:
@rollroyces thanks. Merged current `dev` into your branch: #670 added a `names` key next to the `blocked` line, and both lines are kept. Also folded the two stacked summaries on `blocked_for_entity` into one. Store and server tests pass (480, 0 failed) on a migrated database.

PR #682, opened 2026-09-13
Schema columns carry PK/FK on Trino (#502 cut 3)

rollroyces:
## Summary

#502 cut 3 — Trino / Presto / Starburst `fetch_schema` reads single-column primary keys and foreign keys from the catalog's `information_schema`. Same shape as the MySQL cut (#678); wiremock-tested in CI without external infrastructure.

Single file changed in spirit (`trino.rs`, +352 lines); `snowflake.rs` and `databricks.rs` get a 5-line `schema_row()` call-signature update so they keep compiling until their own cuts land.

## What's in this PR

**`query_engine::trino::keys()`** — module-level helper, mirrors `postgres::keys()` and the MySQL `keys()` shape:

- **Single-column PK** from `information_schema.key_column_usage` joined with `information_schema.table_constraints WHERE constraint_type='PRIMARY KEY'`, with the same composite-PK guard (`count(*) per constraint = 1 AND ordinal_position = 1`).
- **Single-column FK** joined with `referential_constraints` to find the unique-constraint name, then `key_column_usage` again to get the referenced table. Targets come back as `(ref_table_schema, ref_table_name)`.
- **Graceful degradation**: a failing `keys()` returns `Keys::default()` + a `tracing::warn!`, leaving columns without key marks rather than killing `fetch_schema`. The third wiremock test (`a_keys_query_failure_does_not_fail_fetch_schema_trino`) exercises the catalog-admin-revokes-information-schema case.

**`schema_row(row, keys)`** — now takes the keys. Snowflake and Databricks engines pass `&Default::default()` for now; their keys land in cuts 4 and 5.

**`post_paged()` inner function** — extracts the POST + nextUri page-polling loop so both PK and FK queries share it. `headers` is cloned per call; the helper can be invoked twice in the same `keys()` scope.

## Three wiremock tests (no live cluster needed)

1. **`keys_are_marked_when_the_catalog_reports_them_trino`** — mock returns PK and FK rows; assert `tiny.p.id` gets `is_primary_key=true`, `tiny.p.fk` gets `references_table=Some("tiny.p")`, non-key columns stay unmarked.

2. **`a_composite_primary_key_stays_unmarked_trino`** — mock returns empty key sets (the SQL guard on a real cluster filters out the composite PK); assert no column gets `is_primary_key=true`.

3. **`a_keys_query_failure_does_not_fail_fetch_schema_trino`** — PK query returns 500; assert `fetch_schema` still returns columns with `is_primary_key=false, references_table=None`.

The two existing wiremock tests (`follows_next_uri_and_keeps_column_order`, `a_trino_error_page_becomes_an_error`) still pass.

## Trino-specific reality I documented

Trino **does not enforce** PK / FK constraints — they're declarative metadata only (trino-python-client #205, cited in the doc comment). Most catalogs (Hive, Iceberg, Delta, TPC-H) return **empty** `information_schema.table_constraints`, so `keys()` typically returns empty for any read-only BI deployment. This is a feature, not a bug — the graceful-degradation test documents this explicitly.

The PK / FK columns will show up only for catalogs that surface them (e.g. certain Snowflake / Oracle / MySQL connectors when accessed via Trino). For most deployments, `explore_mappings` will get the columns without `[PK]` / `[FK]` markers and behave as before — the schema text is wider without the markers, exactly as #671 specified.

## What I deliberately did NOT add

- **Snowflake / Databricks keys.** The `schema_row(row, keys)` call signature is now `pub(crate)` to accept `&Keys`; Snowflake and Databricks pass `&Default::default()`. Their keys land in cuts 4 and 5 with the same shape.
- **Pg-style permission test.** Trino doesn't have a "CREATE USER" path I can wire up in CI; the `a_keys_query_failure_does_not_fail_fetch_schema_trino` test exercises the equivalent (catalog denies information_schema access → graceful degradation).
- **Live cluster test changes.** The existing `live_tests::a_live_cluster_reads_schema_and_answers` test is unchanged.

## Verification

```
$ cargo check -p utopia-server         # clean
$ cargo clippy -p utopia-server --tests # clean
$ cargo fmt --all --check              # clean
$ UTOPIA_DATABASE_URL=postgres://utopia:***@localhost:1517/utopia \
  cargo test -p utopia-server --bin utopia-server -- --test-threads=1
   271 passed; 0 failed; 1 ignored
```

All three of my new Trino tests pass; the existing Trino wiremock tests still pass; no regressions across the full suite.

## Known limitations in CI: same as #678

The maintainer's `test_db::url()` doc note applies to Trino too: CI's `backend` job doesn't run live integration tests, so `live_tests::a_live_cluster_reads_schema_and_answers` skips silently without `UTOPIA_TEST_TRINO_URL`. The wiremock tests are the CI-passing form — they prove the protocol parsing and the catalog-query shape but not the actual Trino behavior on a real cluster. The `live_tests` module is the right escape hatch for a developer who has a `trinodb/trino` container running.

## Issue

Closes part of #502. After this lands, cuts 4 (Snowflake) and 5 (Databricks) target the same struct shape.

Co-Authored-By: Claude Opus 4.1 <noreply@anthropic.com>
Signed-off-by: Royce <rollroyces@users.noreply.github.com>


WaylandYang:
@rollroyces thanks for keeping the #502 series going. I'm closing this one because Trino can't provide these keys.

Trino's `information_schema` is served by a built-in connector for every catalog, PostgreSQL and MySQL included. It exposes only `columns`, `tables`, `views`, `schemata`, `table_privileges` and the role tables (`InformationSchemaTable` in Trino's source). `table_constraints` and `key_column_usage` resolve to TABLE_NOT_FOUND, so on a real cluster `keys()` always errors and every mount logs a warning without ever marking a key. trino-python-client #205, cited in the doc comment, says the same: PK/FK constraints aren't supported in Trino yet. The wiremock tests pass because the stubs don't look at the SQL. The FK query also selects `ref_table_schema` / `ref_table_name`, which no `information_schema.table_constraints` has.

That's our fault as much as anyone's: the comment in `trino::schema_row` on `dev` says Trino exposes those views. I'm correcting it in a separate PR.

For later cuts, if you want them:
- **Databricks:** Unity Catalog records informational PK/FK in `information_schema.table_constraints` / `key_column_usage` / `referential_constraints`, so this SQL shape fits it better, once the column names are fixed and #678's grouped counting is used.
- **Snowflake:** has no `key_column_usage` and needs `SHOW PRIMARY KEYS` / `SHOW IMPORTED KEYS`.

PR #708, opened 2026-09-14
Models cards say which test and save their note refers to

### LINKED ISSUE #698: Models settings: unclear save and connection-test feedback
## Summary

On **Administration → Models**, the UI makes it difficult to tell whether the current configuration has been saved, which model is being tested, and whether the connection succeeded or failed.

## Observed behavior

1. **Success feedback is barely visible.** In dark mode, `Reachable and authenticated (OK)` and the API key’s `configured` label have very low contrast.

2. **Feedback after changing the API key is ambiguous.** After a successful connection, entering an incorrect key and clicking `Test connection` produces feedback such as `Saved`, without a clear connection result. It is unclear whether the message refers to saving, testing, or a previous operation.

3. **Embedding still shows `Not configured`.** The fields contain `https://api.siliconflow.cn/v1` and `BAAI/bge-m3`, with a masked API key and a `configured` label. However, the card still says `Not configured`, without explaining what is missing or whether the status is current.

4. **Both test buttons appear to respond.** Clicking either `Test connection` button makes both buttons appear to react. It is unclear whether Chat, Embedding, or both are being tested.

## Screenshot

Image

*Low-contrast success feedback and embedding fields populated while the status remains `Not configured`. API keys are masked.*

## Expected behavior

- Clearly distinguish unsaved changes, saved configuration, testing, success, and failure.
- Identify which model and configuration each result applies to, and clear or mark outdated results after edits.
- Display readable success/error messages and explain missing configuration.
- Make button feedback match the actual test scope.

## Environment

- UI version: `v0.1.0`
- Theme: dark
- Provider: SiliconFlow
- Chat model: `deepseek-ai/DeepSeek-V3`
- Embedding model: `BAAI/bge-m3`

These observations concern UI feedback. They do not establish that invalid credentials are accepted or that current inputs are ignored.

aniruddhaadak80:
On Administration → Models, the cards make it hard to tell whether the configuration was saved, which card is being tested, and whether the connection succeeded or failed (#698). Three defects, all still present on `dev`:

1. **Both Test buttons spin together.** The two cards share one `test` mutation, so pressing either one puts both buttons into "Testing…".
2. **Stale conclusions.** A test result (or "Saved") stays pinned to the card after any field is edited — but the endpoint always tests the *saved* configuration, so the note describes a configuration the form no longer shows. The screenshot's "fields populated yet Embedding says Not configured" is this mechanism: the status belonged to the previously saved config.
3. **Inconsistent tones.** Success renders in neutral `text-accent` on both cards; chat errors render `text-danger` while embed errors render body-tone `text-ink-2` — unlike the Data sources section in the same file, which uses `text-ok`/`text-danger`.

A transport-level test failure (request never reaches the endpoint) additionally showed nothing at all: the note silently fell back to the old save state.

## What changed (`web/src/pages/Settings.tsx`)

- `testCard` state: "Testing…" lights up only on the card whose button was pressed. Results still return to their respective cards (one endpoint tests both).
- Freshness resets: editing any field (or applying a preset) clears the old test result, save confirmations and errors; saving a card expires the previous test result (it described the pre-save config); starting a test yields the floor to the new round.
- `modelCardStatus` (pure, exported): one precedence table for both cards — fresh test result (including a transport failure) beats save error beats "Saved" beats silence — returning the `text-ok`/`text-danger` tone with the note.
- No new UI strings; no API changes.

Fixes #698

## Verification

- New `web/src/pages/Settings.test.ts` (5 tests): fresh result beats stale "Saved", failed test beats save state, transport failure surfaces instead of going quiet, save error/confirmation preserved when untested, silence when idle. Each assertion contradicts the previous rendering (old tones were `text-accent`/`text-ink-2`, old transport failures fell back to save state).
- `pnpm vitest run`: 11 files, 56 tests, all pass (51 pre-existing + 5 new).
- `pnpm typecheck` (`tsc --noEmit`), `pnpm guard` (style guard: 80 files compliant), and `pnpm build` (style-guard + typecheck + vite) all green locally — i.e. the full `web` CI job reproduced.

## Notes / trade-offs

- The backend's raw `"Not configured"`/`"Empty response"` strings are still displayed verbatim on failure; localizing them would mean matching on backend message text across the API boundary, which is brittle — leaving that as a possible follow-up.
- The contrast complaint in #698 dates from v0.1.0; the token system since then gives both themes readable tones, so this PR aligns the cards to the existing `ok`/`danger` semantic tones rather than introducing new ones.

## After review (6b1ecd93)

Test checks the saved configuration, so a card with unsaved edits now says "Unsaved changes. Save this card to test them." and disables its Test button until saved. This covers #698 items 2 and 3: an edited key tested against the old one, and filled but unsaved fields reported as "Not configured". Checked in the browser against a real endpoint:
- only the pressed card shows "Testing…";
- editing the chat key marks that card unsaved and clears the embedding card's stale result;
- saving a wrong key and testing shows the 401 in red on the chat card and the dimension in green on the embedding card.


WaylandYang:
@aniruddhaadak80 thanks, this makes the cards much clearer. I checked it in the browser against a real endpoint:
- Test on one card spins only that card.
- Results come back green and red on each card, including a 401 from a wrong key after saving.
- Editing a field clears stale results.

One gap from #698 remained, so I pushed a commit to your branch before merging (6b1ecd93): **Test always checks the *saved* configuration.** Items 2 and 3 in the issue are exactly that case. You type a new key, press Test and see the old key's "Reachable"; or the embedding fields are filled but unsaved and the card says "Not configured".

Now:
- a card with unsaved edits says "Unsaved changes. Save this card to test them." (en and zh) and its Test button is disabled until saved;
- a successful save clears that state.

`modelCardStatus` takes a `dirty` flag, with one more test. Vitest 57/57 and `pnpm build` pass.

PR #711, opened 2026-09-15
An entity panel lists the entity's names

WaylandYang:
The entity panel shows an entity's names, and a wrong one can be removed.

## Why

Since #670 (0041) every name an entity carries is a `known_as` fact, and `GET /kbs/{id}/entities/{entity}` returns them under `names`: the shown name, abbreviations and former names, each with sources and validity. The panel never rendered them. Name facts are also kept out of the facts list on purpose, so they do not pass for relations. The result was that nothing in the interface showed what an entity is also called.

The extraction audit behind #710 found that this is where wrong merges start. "OpenAI's", "ChatGPT apps" and "… dated 2020-xx-xx" were recorded as aliases. "deep reinforcement learning" was recorded as an alias of "reinforcement learning", the pair the governance bench keeps merging. None of these could be seen until they had caused a merge.

## Change

- **`api.ts`:** `NameView` and `names` on `entityDetail`.
- **Entity panel, Relations view:** a Names section above the From / To sections, shown only when the entity has more than one name. One row per name, with:
  - "shown name" on the canonical one;
  - "until <date>" on a former name, dimmed like past facts;
  - the source count;
  - a hover-revealed Remove on every name but the shown one.
- **Remove** calls the existing fact rejection (`POST /kbs/{id}/facts/{fact_id}/reject`). It is audited and appears in History. Renaming stays in the edit dialog.
- **Copy:** English and Chinese strings.

No server change.

## Checks

- `tsc --noEmit`, `pnpm guard` (style guard) and `vitest` (57 tests) pass.
- To look at it: the dev server on 5204 proxies to the e2e clone on 1523. The base "identity pn2-a forward" has 东海海洋研究所 with four names after the rename merge.

Waiting for a look before merge.


WaylandYang:
Looked at it in the browser on the identity bench base (东海海洋研究所, 4 names): the Names section lists the shown name first, then the others with source counts, and Remove appears on hover for every name except the shown one. No console errors beyond the login redirect. `pnpm build` (style guard and tsc) passes, and vitest 57/57 on a merge with current `dev`. Merging.

Worth noting what it surfaced: 澜海研究院财务部 and 澜海研究院理事会 are recorded as names of 东海海洋研究所. Those are sub-organisations, the kind of wrong alias this panel exists to catch. The known issue that a rename merge keeps the old canonical name is on the server side (`merge_entities`), not in this change.

PR #713, opened 2026-09-15
Add 100k benchmark first cut (cold SELECT scenario)

rollroyces:
## Summary

Roadmap item: **Enterprise — benchmarks at 100k documents** (README §Roadmap). First cut: a Rust integration-test benchmark against a 100k-document synthetic corpus, scenario 1 (cold `SELECT ... WHERE`) wired up.

## What's in this PR

- **`crates/utopia-store/tests/bench_100k.rs`** (148 lines) — integration test that:
  - Builds a synthetic `bench_100k_docs` table (BIGSERIAL id, kb_id, kind, name, body, n, created_at) populated with 100 000 rows in 20 batches of 5 000.
  - Runs 30 iterations of a single `SELECT id, body, n FROM bench_100k_docs WHERE kind = $1 AND kb_id = $2 AND n > $3`, timed with `std::time::Instant`.
  - Prints p50 / p95 / p99 / mean / throughput as a markdown table.
  - Deterministic seed — same hardware, same numbers, two runs.
  - **Drops the table at the end.** CI leaves nothing behind.

- **`.roadmap-proposals/benchmarks-100k.md`** — design with 5 open questions.

## Output (live run)

```
| scenario     | n  | p50 (ms) | p95 (ms) | p99 (ms) | mean (ms) | throughput |
|--------------|----|----------|----------|----------|-----------|------------|
| cold_select  | 30 |    11.45 |    23.95 |    24.68 |    14.33  |   69.8 q/s |

rows_returned_total = 237490
test cold_select_at_100k_rows ... ok
```

## Why integration test, not `benches/foo.rs`

- House style is integration tests under `crates/*/tests/`; `scripts/bench/*.mjs` is the existing JS measurement surface. Adding a Rust bench layer would be a third pattern.
- No criterion / no divan. `grep -r 'criterion\|divan' --include=Cargo.toml` returns nothing in the workspace. An integration test using `std::time::Instant` directly is 30 lines, not 300 of framework boilerplate.
- The pattern matches the existing `vector_index.rs:181` measurement and the skip rule in `crates/utopia-store/src/test_db.rs:14`.

## Skip rule (matches `test_db.rs:14`)

- `UTOPIA_DATABASE_URL` unset → bench prints "skipping" and returns `Ok(())`. Local `cargo test` without a DB still works.
- `UTOPIA_TEST_REQUIRE_DB=1` and `UTOPIA_DATABASE_URL` unset → bench panics with "this run must not skip database-backed tests". This is the CI signal.
- The existing `migrations` CI job already runs store-integration tests against a real Postgres. No new workflow needed.

## Scenarios not yet landed

1. **Cold SELECT with WHERE** ← this PR.
2. **Warm cache** (re-run scenario 1 with `pg_prewarm`). Validates cold-vs-warm gap.
3. **Write-heavy bulk INSERT.** Times a 5 000-row batch — the same shape the corpus generator uses.
4. **Extraction-throughput.** Times how long it takes to extract facts from a sample of 1 000 documents end-to-end. The closest the suite gets to "100k documents" in the roadmap item.

Each is a separate PR once (1) lands.

## Open questions

Full text in the design doc. In priority order:

1. Bench format: integration test (this proposal) vs. `benches/foo.rs` with criterion.
2. Where the result lands: print to stdout (this proposal) vs. `docs/benchmarks/<date>.md` vs. PR-comment CI.
3. What "document" means: `documents` row, `chunks` row, `facts` row, or source-side doc?
4. Ship all 4 scenarios at once, or one at a time?
5. Is the bench a merge gate, or informational? Decision 0035's precedent is informational.

## What this PR does NOT do

- No scenarios 2–4 (gated on Q4).
- No `docs/benchmarks/` report file (gated on Q2).
- No CI workflow change — the existing `migrations` job picks this up via the integration-test pattern.

## Verification

- Local: `UTOPIA_TEST_REQUIRE_DB=1 UTOPIA_DATABASE_URL=postgres://utopia:utopia@127.0.0.1:1517/utopia_test cargo test -p utopia-store --test bench_100k -- --nocapture` → 1 passed in 1.08s.
- CI: the `migrations` job runs store-integration tests against a real Postgres; this bench will run there.

## Checklist

- [x] DCO sign-off (`Signed-off-by:` in commit)
- [x] `cargo fmt --all` clean
- [x] `cargo check --tests -p utopia-store --test bench_100k` clean
- [x] `cargo test --workspace` 715 passed, 0 failed
- [x] Existing store tests unaffected (no changes to existing test files)
- [x] No new deps, no migration, no workflow change


WaylandYang:
@rollroyces thanks for picking up the 100k item. I ran the test and checked it against a real Postgres, and I don't think it can merge in this shape. It measures Postgres, not Utopia.

**What it measures today**
- **No real tables or code.** It never touches `documents`, `chunks`, `entities` or `facts`, and never calls Utopia code. The only index on `bench_100k_docs` is the primary key, so each iteration is a sequential scan of 100k rows. The roadmap item asks whether the product stays fast with 100k documents in a base, and this can't answer that.
- **It isn't cold.** `EXPLAIN (ANALYZE, BUFFERS)` right after the load shows `shared hit=1031` and zero reads, even for the discarded first query. The rows were just written, so every page is in shared buffers.
- **A third of the rows can't be queried.** `id.wrapping_mul(0x9E37_79B9_7F4A_7C15_u64 as i64) % 4 + 1` overflows. Rust's `%` keeps the sign, so `kb_id` takes the values -2, -1 and 0 (12.5k rows each) as well as 1–4, and `kb_id = 1` holds 25k. The queries only bind 1–4, so 37.5% of the table is never read.
- **The description names another query.** It says `SELECT id, body, n … AND n > $3`; the code runs `SELECT id, name … WHERE kb_id = $1 AND kind = $2`. The table you posted does come from this code (237490 rows total matches).
- **It runs in CI.** It lives with the store integration tests, so the `migrations` job loads 100k rows on every PR (1.8 s in the log) and prints numbers nobody reads. Timings on shared runners aren't comparable anyway.
- **A failed load leaves the table behind.** If loading fails after `CREATE TABLE`, `teardown` never runs, and the next run on the same database fails with "already exists".
- **The design doc directory again.** As noted on #691, `.roadmap-proposals/` isn't a convention here; reasoning goes in `docs/decisions`, or in the PR description while a design is open.

**What a useful 100k benchmark looks like**
- **Data:** a generator that fills the real tables with a realistic base. For example: 100k documents, a few hundred thousand chunks, facts in the millions with merges and corrections in their history, and a few hub entities with very many facts. Writing rows directly is fine (no model calls), but the shapes and distributions should be real.
- **Timed operations are the product's own paths, called through `utopia_store`:**
  - `graph::search_entities`
  - entity detail and timeline
  - hybrid retrieval (BM25 and vector)
  - replay with `as_of` on the record axis
  - `temporal::reconcile_new_fact` on a hub entity
  - paging a review queue
  - deleting a document cited across many timelines
- **How it runs:** manually, behind an env var or `#[ignore]`, not in CI. Results go into a dated report in the repository, with the machine and Postgres settings, so runs can be compared.

**Your five questions**
1. **Format:** an integration test with `Instant` is fine; no criterion.
2. **Output:** a report file in the repository, not stdout.
3. **What "document" means:** a `documents` row, together with its chunks and facts.
4. **Scenarios:** one at a time is fine, but the first cut should already run on the real tables and a real product path.
5. **Merge gate:** informational only, like 0035.

I'll leave this open in case you want to rework it on this branch; otherwise close it and open a new one in the shape above.

WaylandYang:
Thanks for this and for the design doc. Answers to the five questions:

1. Integration test, no new dependencies: yes.
2. Write the result to `docs/benchmarks/<date>.md` rather than stdout, so a number from one month can be compared with the next.
3. A "document" is a `documents` row together with its `chunks` and `facts` rows; the corpus should look like the product's ledger, not a table of its own.
4. One scenario at a time: yes.
5. Informational, as 0035: yes.

That answer to 3 changes what the first scenario has to be. As it stands, the bench times `SELECT ... WHERE` on a synthetic `bench_100k_docs` table, which tells us Postgres can filter 100k rows but nothing about utopia: none of our tables, indexes or read paths are touched. Please rework scenario 1 to populate `documents`, `chunks` and `facts` for one base (the store's own write functions are fine, or a fixture generator that uses them) and time one real read path against them, for example the graph read a base opens with (`graph::facts_hold_at` with `at` and `as_of`) or the entity search. With that, the number means something and the later scenarios have a corpus to grow. The harness, the skip rule and the percentile table can stay exactly as they are.

rollroyces:
Superseded by #782 (design doc) and the upcoming code cut. The previous shape measured Postgres, not Utopia — see the new design at docs/design/100k-cold-select.md. Closing this so review bandwidth is on one thing at a time.

PR #715, opened 2026-09-15
The extraction progress counts each document once, not once per pipeline

zhongqishang:
## Problem

The Library progress bar read its two numbers from two different pipelines. The numerator was `count(*) FILTER (WHERE status = 'ready')` — the ingest pipeline — while the denominator was `ready + extracting`, and `extracting` counts `graph_status IN ('queued', 'extracting')` — the extraction pipeline. A document that had finished ingesting while its extraction was still running was counted in both, so one document could appear twice.

On the library that reported it (`01a09ed7`), the two live documents both had `status = 'ready'`, with `graph_status` `done` and `extracting`. That gave `ready = 2`, `extracting = 2`, and the bar read `抽取中 · 2 / 4` for two files.

The numbers themselves are computed fresh on every request (`count(*) FILTER`, no stored counter), so nothing was stale — only the two columns being added together.

## Change

- `DocumentPage` gains `done` — `count(*) FILTER (WHERE graph_status = 'done')`.
- The bar takes both numbers from the extraction column alone: `done` as the numerator, `total = done + extracting`.
- `ready` keeps its old meaning. The re-extract confirmation still uses it to say how many documents the action will queue.

## Checks

- `cd web && pnpm typecheck` passes.
- The bar was checked against the live library after deploying the build: the numbers now follow that library's `graph_status` distribution rather than exceeding its document count.

## After review (70670c20)

- **Test:** `extraction_progress_counts_a_document_once` (DB-backed, in `utopia-store`) seeds one document each that is done, extracting, queued, failed, and deleted while extracting. It checks `ready = 4`, `done = 1`, `extracting = 2` (queued counts as extracting) and `failed = 1`, with the deleted document in none of them. The bar reads 1 / 3; the old formula would have read 4 / 6.
- **The old numerator was `ready`**, because `total - extracting = ready`, so any document still being extracted counted as finished. A library whose only document is extracting read 1 / 2.


WaylandYang:
@zhongqishang thanks, the fix is right. I merged current `dev` into your branch and added a DB-backed test covering live, queued, failed and deleted documents (70670c20). Merging.

PR #727, opened 2026-09-16
Report get_document read failures as tool errors

Maya-Kid:
`get_document` currently reports a failed document lookup as a successful "not found" result. If the lookup succeeds but reading its chunks fails, it instead returns a successful empty document. Both can make a caller treat a storage failure as an answer about the document.

Handle both database errors explicitly, log the underlying error server-side, and return `isError: true` with "Could not read the document." This extends the read-error handling in #604 to `get_document`. Missing, deleted and foreign-KB documents retain the same indistinguishable response; genuinely empty documents still succeed. No dependencies, migrations or structured output fields change.

Regression coverage uses a closed pool for lookup failure and a connection-local search path with an incomplete `chunks` view for chunk-read failure. The latter verifies that the document lookup succeeds first and does not alter shared tables. Normal, empty, missing, deleted and foreign-KB reads are also covered through the authenticated MCP handler.

Validation:

- Before the fix: the two failure regressions fail; the normal/missing-document test passes.
- After the fix: all 10 MCP tests pass against PostgreSQL with `UTOPIA_TEST_REQUIRE_DB=1`.
- Rust 1.98.1 (matching upstream CI): formatting, workspace Clippy and workspace build pass; `cargo test --workspace` passes (826 passed, 1 ignored, database-gated cases skipped as in the backend CI job).
- Frontend: frozen-lockfile install, all 63 tests and production build pass with pnpm 10.2.1.

Additional runs against the SSH-forwarded database were not fully green: two pipeline timing failures (`a_failed_batch_does_not_strand_the_document` and `the_embedding_gate_is_never_held_beyond_its_ceiling`) also reproduce on unmodified `bb564694`; the separate store suite stopped at the unchanged `a_queued_job_wakes_an_idle_worker` timing assertion (433 ms versus its 300 ms bound). The database-backed MCP regression suite passes independently.


WaylandYang:
Merged — thank you. Verified locally too: the two regressions fail on dev with only `tools.rs` reverted, and pass with the fix; the full utopia-server suite (295 tests), workspace clippy and rustfmt are clean. The timing failures you saw did not reproduce here.

PR #728, opened 2026-09-16
Apply the retrieval result limit after visibility filtering

Maya-Kid:
An `as_of` search can return no results even when an eligible hit has already been recalled. `hybrid` currently applies the final `top_k` limit before `chunks_by_ids` filters by record time and document visibility. Six high-ranking May chunks can therefore crowd out a January chunk when querying March with a limit of six; increasing the limit to seven makes that same January chunk appear.

Fuse the bounded channel candidates, apply the existing database filters, then truncate the eligible results. Per-channel recall remains capped at 24; this fetches at most 48 candidate chunks across the two channels. It preserves RRF ordering and does not add historical versions to the current-only BM25 index.

The regression verifies that all seven candidates are present in the index, that the current query retains its ranking and limit, and that historical limits of six and seven both return the eligible chunk. Run the retrieval suite in the PostgreSQL CI job with `UTOPIA_TEST_REQUIRE_DB=1` so these cases execute rather than silently skip. The fixture now applies migrations for standalone test runs.

Validation:

- With the new regression on unchanged production code: fails with `[]` instead of the January chunk; the current-query assertions pass.
- With the fix: all 6 retrieval tests pass against PostgreSQL 16 / pgvector 0.8.6, with database tests required.
- All 10 MCP database tests pass with database tests required.
- Rust 1.98.1: formatting, workspace Clippy with warnings denied, and workspace build pass.
- Workspace tests in the backend CI configuration (without a database URL): 827 reported passed, 1 ignored; database-gated cases skip in this run and were exercised separately above for retrieval and MCP.
- Frontend: pnpm 10.2.1 frozen-lockfile install, 63 tests, and production build pass.


WaylandYang:
Merged — thank you. Verified locally: the new regression returns `[]` for a limit of six on dev and the eligible chunk with the fix; utopia-server (296 tests), clippy and rustfmt are clean. Good catch on quoting the CI command — the unquoted `retrieval::` kept GitHub from creating a run at all.

Issue #714, opened 2026-09-15
Upload time is used as the document date in extraction


WaylandYang:
An undated upload stores `doc_time = now()` with `doc_time_source = 'upload_time'` (`crates/utopia-store/src/documents.rs`). That time is when the ledger learned of the document, the transaction axis. Only `crates/utopia-store/src/temporal.rs` filters on `doc_time_source`; extraction uses `doc.doc_time` as if it were the document's own date:

- `crates/utopia-server/src/extraction.rs` passes it to the prompt, where `crates/utopia-extract/src/lib.rs` tells the model "Document date: {t}. Resolve relative time expressions … using it as the reference." A "today" or "last year" in an undated document resolves against the upload day.
- `validity_of` sets `attested_at: doc_time`, so facts from an undated document are attested at the upload moment, and open-started facts hold from it.
- name facts are recorded with `doc.doc_time` as their time.

Seen on 2026-09-15: every document in the recall-bench KB has `doc_time_source = 'upload_time'` and `doc_time` equal to the upload day; a prototype that passed it as "dated" produced "the 2024 statistical bulletin — dated → 2026-09-14".

Expected: a `doc_time` whose source is `upload_time` is treated as no document date in extraction (prompt says the date is unknown, no relative resolution against it, `attested_at` not taken from it). The document's own date should come from the source or from its content.

AaronProbha18:
Hi @WaylandYang,
I’d like to work on this issue.

## Plan
The bug is that undated uploads receive an `upload_time` timestamp, and `extraction.rs` currently interprets that timestamp as the document’s actual date. That affects both relative-date reasoning in the extraction prompt and `attested_at` on extracted facts. I’ll introduce one gated “document-own-date” value: upload-time documents become dateless, while genuinely dated documents retain the existing behavior. The same value will feed both consumers so the two paths cannot drift apart.

The implementation will start by confirming the current call sites, then make the server-side change, add coverage for both upload-time and genuinely dated documents, and run the existing workspace tests. Any matching decision record under `docs/decisions/` will also be updated if present.

## Changes
* `crates/utopia-server/src/extraction.rs` — centralize the `doc_time_source` check and use the resulting value for both prompt construction and `Validity.attested_at`. Since one `Validity` binding is reused across fact/name-fact inserts, this fixes all downstream writes without touching each call site.
* `crates/utopia-extract/src/lib.rs` — no change; its existing `None` handling already produces the correct dateless prompt behavior.
[...]

PR #730, opened 2026-09-16
fix: undated uploads using extraction-time recording anchors

AaronProbha18:
## What changed
`documents.create` stores `doc_time = now()` with `doc_time_source = 'upload_time'`
for an undated upload — a record-axis timestamp standing in a world-axis column.
`extraction.rs` read `doc.doc_time` at five sites and never looked at
`doc_time_source`, so an upload-time value was used both as "the document's own
date" in the prompt and as the recording anchor for every fact and name fact
extracted from it. A memo uploaded today that says "last year" resolved to last
year *from today*, and every fact it produced anchored on today.

The two sites want different things, so they stop sharing a value:

- **The prompt** genuinely wants no date. `prompt_doc_time` is `None` when
  `doc_time_source == "upload_time"`; `build_messages` already handles that.
- **The anchor** must never be absent. `attest_earlier` no-ops on `None` ("now
  can't be earlier than any existing evidence"), so a null anchor would leave
  `close_with_unknown_end` nothing to close against and an ended relation would
  go on reading as still holding. `recording_anchor` falls back to this run's
  `Utc::now()` — 0022 section 3's "the moment of recording", now meaning the
  moment of extraction rather than of upload. Computed once per run so every
  fact from one document shares an anchor.

All four anchor sites take `Some(recording_anchor)`; no site reads
`doc.doc_time` directly any more. `graph.rs`, `temporal.rs` and `names.rs` are
unchanged — their logic was already right given a real anchor; the bug was
entirely in what extraction handed them. No schema change.

Dated documents (`doc_time_source` of `content` or `source`) behave exactly as
before.

0022 section 3 gets a dated revision note.


WaylandYang (changes requested):
Thanks for picking this up. The prompt half is exactly what #714 asked for. The anchor half needs another pass, and a few requirements from the issue thread are still missing.

**1. Prompt: keep it.** `prompt_doc_time` is `None` for `upload_time`, so a "last year" in an undated upload no longer resolves against the upload day. 👍

**2. Anchor: please drop the `Utc::now()` fallback.** Extraction time is still a transaction-axis timestamp, the same kind of value #714 is about. It just comes from a different clock.
- In the normal flow, extraction runs minutes after upload, so the anchors barely change.
- On re-extraction, the anchor drifts later.
- The direction we've taken is recorded in 0045 (#724): attested time stays **null** for an undated document, and closing a timeline no longer depends on it. That work lands with 0045's cuts.
- So please leave the four anchor sites as they are on dev for now, and remove the 0022 revision note (it would record the opposite direction).

**3. Tests.** Please add the ones asked for in the thread, scoped to what this PR now changes:
- an undated upload (`doc_time_source = 'upload_time'`) builds a prompt with no document date;
- a dated document (`content` / `source`) still gets its date in the prompt.

**4. CI.**
- `cargo fmt --all --check` fails on the `validity_of` call.
- The commit has no `Signed-off-by`, so DCO is red. `git commit --amend -s` and a force-push will fix it.

With 1–4 done this is a small, safe merge.

AaronProbha18:
All four done — thanks, the 0045 pointer is the part I'd missed.

**1.** Kept, and pulled out into `prompt_doc_date(doc_time, doc_time_source)`
so it's testable without a document. `upload_time` → `None`, everything else →
the formatted date. `build_messages` already says the date is unknown on `None`,
so nothing changed there.

**2.** Dropped. The four anchor sites are back to `doc.doc_time` exactly as on
dev, and the 0022 revision note is gone (`git diff dev -- docs/` is empty).
You're right that extraction time is the same kind of value — it's a different
clock, not a different axis, and on re-extraction it drifts later for no reason
the ledger can defend. This PR is now the prompt half only.

**3.** Two unit tests beside it:
- `an_undated_upload_gives_the_prompt_no_date` — `upload_time` with a real
  `doc_time` yields no date;
- `a_dated_document_still_carries_its_date_into_the_prompt` — `content` and
  `source` both keep `2024-02-20`, and a genuinely dateless document still
  gives `None`.

Extracting the helper was mostly so these could be plain unit tests rather than
standing up a document and an LLM for a one-line branch.

**4.** The `validity_of` call is back to its dev form, so the fmt failure goes
with it. Commit is signed off — I squashed the DCO remediation commit into the
real one rather than leaving both, so the branch is a single commit with one
`Signed-off-by`. Force-pushed.

WaylandYang:
Thank you for this. It is superseded by #740 (0045): the open contract no longer sends any document date to the model, and migration 0064 sets `doc_time` to NULL with source `none` for every document whose only date was the upload time or the file modification time, so #714 is closed there at the ledger rather than at the prompt. The typed path this patch edits (`extraction::run`) is deleted in #736. Closing in favour of that stack.

PR #732, opened 2026-09-16
Closing a fact clears it from the unconfirmed queue

zhongqishang:
## Problem

The "unconfirmed" queue on the Review page is meant to list facts that still hold, so that closing one actually clears it. Closing goes through retract-and-rewrite (`temporal::close_superseded`), and the superseding row copies the evidence. The row that came back therefore still matched the queue's criteria: entering `2021-01` and pressing close left the row in the list and the badge unchanged, and pressing close a second time hit `close_fact`'s "only an open interval can be closed" check and returned `NotFound`.

The criteria were `invalidated_at IS NULL` plus "every evidence chunk superseded". The comment above `stale_facts` calls these *current* facts, but the open upper bound was never tested, so the close path could not succeed. 0022 already states the rule: `valid_to IS NULL` beside `valid_to_precision IS NULL` means "still holds", while `'unknown'` is not an open end.

## Change

- `review::UNCONFIRMED_FACT` gains `f.valid_to IS NULL AND f.valid_to_precision IS NULL`. It is shared by `counts` and by `review_summary`'s waiting and health blocks, so the badge, the overview and the list stay on one criterion.
- `graph::stale_facts` — the list — gains the same clause.
- New DB-backed test `closing_an_unconfirmed_fact_clears_the_queue.rs`. A fact whose only evidence sits on a superseded chunk is listed; after `close_superseded` it leaves the queue and `unconfirmed` drops to 0, while the evidence is still copied onto the corrected row — the criterion is the bound, not the evidence disappearing. A second case covers `valid_to_precision = 'unknown'`, which `valid_to IS NULL` alone would have kept in the queue.

## Checks

- `cargo fmt --all --check` clean.
- The two new tests and the existing `a_review_has_a_summary` (which exercises the same WHERE through `counts`, `waiting` and `health`) pass against a real Postgres with `UTOPIA_TEST_REQUIRE_DB=1`.
- `close_superseded`, `stale_facts` and `UNCONFIRMED_FACT` are identical on this base and on the deployed build the report came from, so the fix carries over as written.


WaylandYang:
Merged — thank you, nice catch. Verified locally that both new tests fail on dev with only `graph.rs` / `review.rs` reverted and pass with the fix; `a_review_has_a_summary` still passes. I pushed one small commit before merging: the tests now also delete their organization and workspace (deleting a knowledge base doesn't cascade up), so repeated runs don't leave rows behind in a shared test database.

PR #733, opened 2026-09-16
Report rule read failures as tool errors

Maya-Kid:
When database reads fail, `list_rules` and `rule_matches` already return failure text, but their `ToolResult` defaults to success. MCP therefore returns `isError: false` beside "Could not read the rules." or "Could not read what that rule marks." Clients that inspect the flag cannot distinguish those failures from successful reads.

Mark just the two database-error branches with `.error()`. Extend the existing failed-read regression with both tools, and add authenticated MCP coverage for normal rule listings, matches, and successful empty results. Response text and successful-read behavior stay the same. The existing PostgreSQL CI job already runs this MCP suite.

Validation:

- New tests on unchanged production code: 10 pass and the failed-read regression fails on `isError: false` for `list_rules`.
- With the fix: all 11 MCP tests pass against PostgreSQL 16 / pgvector 0.8.6, with `UTOPIA_TEST_REQUIRE_DB=1`; both failed-read branches are covered.
- Rust 1.98.1: workspace Clippy with warnings denied and formatting pass. A separate build cache was used for the final validation of this branch.
- Workspace tests without a database URL: 828 reported passed, 1 ignored. Database-gated cases skip in that run; the 11 MCP cases above were separately run with a database required.
- Frontend: frozen-lockfile install with pnpm 10.2.1, all 63 tests, and production build pass. The two submissions have identical, unchanged frontend trees, so this check was shared.
- The unchanged CI workflow passes YAML parsing and actionlint.
- Workspace build passes.


WaylandYang:
Merged — thank you. Verified locally: the failed-read regression goes red on dev with only `tools.rs` reverted and all 11 MCP tests pass with the fix.

PR #734, opened 2026-09-16
Reuse the vector build connection for index status checks

Maya-Kid:
`vector_index::build` acquires a connection for its session settings and concurrent DDL, then calls `status(pool, ...)`, which acquires a second connection. With a supported two-connection pool, another request holding the other connection is enough for the build to fail with `PoolTimedOut`, even though the build already owns a usable connection. Two builds can also exhaust the pool this way.

Reuse the held connection for the status query through a private executor-based helper. The public `status` interface, query, index DDL, and session-setting cleanup stay the same.

The database regression holds one connection in a pool of two to make the condition deterministic. It checks creation and reuse of valid indexes for both chunks and entity profiles. On unchanged production code it fails with a pool acquisition timeout; with the fix it passes. The timeout bounds the failed nested acquisition, rather than imposing a duration limit on index construction. The existing PostgreSQL CI job already runs this store test.

This addresses nested acquisition under pool pressure; it does not claim that default 32-connection deployments commonly hit the problem or that all index concurrency issues are resolved.

Validation:

- New regression: fails on the dev implementation, passes with the fix against PostgreSQL 16 / pgvector 0.8.6, with `UTOPIA_TEST_REQUIRE_DB=1`.
- Existing index queue/build regression also passes.
- Rust 1.98.1: formatting, workspace Clippy with warnings denied, and workspace build pass.
- Workspace tests without a database URL: 828 reported passed, 1 ignored. Database-gated cases skip in that run; the index regressions above were separately run with a database required.
- Broader store database suite: stopped before completion; full database validation is pending CI. Completed test groups reported 249 passes and zero failures; this is not a claim that the complete suite passed. The final result remains for CI to establish.
- Frontend: pnpm 10.2.1 frozen-lockfile install, 63 tests and production build pass (shared check of the identical, unchanged frontend tree).
- The unchanged CI workflow passes YAML parsing and actionlint.


WaylandYang:
Merged — thank you. Verified locally: the new regression times out on dev with only `vector_index.rs` reverted and passes with the fix, alongside the existing queue/build test.

Issue #729, opened 2026-09-16
A statement keeps the document's words


WaylandYang:
Cut 1 of 0044 (#710): the open graph lands in the ledger, and extraction can write only that layer. Follows the sequence in #725 (queues come after cuts 1 and 2) and the time record 0045 (#724), whose interpretation columns wait for a later cut.

## What the ledger gets

An open statement is what a document says, in its own words: a subject, a relation phrase as written, an object entity or a literal value, qualifiers keyed by the document's role words, the time words it mentions, and provenance to the chunk. Today an unbound relation is representable (0010: `predicate_id` NULL, phrase on `fact_evidence.proposed_predicate`), but it dedups badly (NULL-blind key, every observation inserts a row), qualifiers must be keyed by an ontology attribute (0037), a described thing cannot be an entity (`canonical_name` and the `known_as` fact are mandatory), time exists only as computed dates, and nothing marks the layer.

Migration `0061_a_statement_keeps_the_documents_words.sql` (0060 is taken by #721). Additive, no backfill; there is no released version.

```sql
ALTER TABLE knowledge_bases ADD COLUMN open_extraction BOOLEAN NOT NULL DEFAULT FALSE;

ALTER TABLE facts
  ADD COLUMN layer TEXT NOT NULL DEFAULT 'typed' CHECK (layer IN ('typed', 'open')),
  ADD COLUMN phrase TEXT,
  ADD CONSTRAINT facts_open_statement_shape
    CHECK (layer <> 'open' OR (phrase IS NOT NULL AND predicate_id IS NULL));
CREATE INDEX facts_open_by_subject ON facts (kb_id, subject_id, phrase)
  WHERE layer = 'open' AND invalidated_at IS NULL;

ALTER TABLE fact_evidence ADD COLUMN quote_start INT, ADD COLUMN quote_end INT;  -- char offsets into chunks.text; NULL = not located

ALTER TABLE entities ADD COLUMN description TEXT;  -- a described, unnamed thing

CREATE TABLE statement_qualifiers (
  fact_id   UUID NOT NULL REFERENCES facts (id) ON DELETE CASCADE,
  role      TEXT NOT NULL,                         -- the document's word
  value     JSONB,
  entity_id UUID REFERENCES entities (id) ON DELETE CASCADE,
  PRIMARY KEY (fact_id, role),
  CHECK ((value IS NOT NULL) <> (entity_id IS NOT NULL))
);

CREATE TABLE time_mentions (
[...]

PR #743, opened 2026-09-17
A statement reads on its own and a thing is named by a term

WaylandYang:
Stacked on #741. Two contract rules from the pitfalls list of `docs/design/prior-work.md` (items 1 and 6), and a second dimension on the judge.

## What lands

**A phrase reads on its own.** The relation phrase is the verb with the words that belong to it, so that subject, phrase and object read as a sentence without the document; never a bare verb cut from a longer verb phrase ("went into partnership with", not "went"). Several verbs sharing one object are one statement; one verb with several objects is one statement per object. The object is what the verb acts on, and a described thing that is the subject or the object is listed like any other; where, how, why, with what and for whom go in qualifiers under the passage's own role word. A named thing mentioned in a qualifier is still listed and the qualifier names it; a generic phrase that appears in a qualifier alone stays words there and is not a thing.

**A thing is named by a proper name or a fixed term.** `named = 1` for the name of one particular person, organization, product, place, document, law or event, or for a term that means the same thing in any document (a disease, a drug, an industry, a product category, an indicator). `named = 0` for a role or a generic phrase whose referent the passage decides ("the company", "patients", "各部门"), however particular it is there. Named things resolve across documents by name; described things stay in their document. This is a term gate, not a proper-name gate: on the batch below the model already kept 企业, 各地区 and patients described; what changed is that symptoms, procedures and product categories (nausea, liver biopsy, 乙烯) are named and merge across documents by design.

**The judge reports `alone`.** `judge_open.mjs` asks, beside stated / misworded / not stated, whether "subject —phrase→ object" reads as a proposition by itself. It is reported separately so the 2% threshold on not stated stays comparable.

No server check changed: these are contract rules, and the shape checks look at structure, not vocabulary.

## Measured

Same 24 short documents (three 8-Ks, nine FDA release paragraphs, seven statistics-bulletin paragraphs, two trade-in notice paragraphs, the harbor text), same model at temperature 0, the contract of #741 against this one; the judge is a second model:

| 24 short documents | #741 | this change |
|---|---|---|
| statements | 594 | 538 |
| not stated | 0.5% | 0.7% |
| misworded | 7.4% | 5.8% |
| reads alone = false | 11.1% | 9.1% |
| named / described things | 136 / 325 | 158 / 303 |
| qualifiers | 310 | 309 |

Per document the readability moved where the rule aims: Journavx 46% → 0%, Qfitlia 13% → 0%, the 8-K exhibit 13% → 2%, the statistics bulletins 6% → 0%, the trade-in notice 38% → 34%. What is left as "does not read alone" is three other classes: table rows whose label needs its section (vote counts, 住宅 under 商品房销售面积), the implicit subject of directive sentences (the model takes the section heading as subject), and the unfolded-description statements of rule 4 (patients —with→ underlying adrenal insufficiency), which raised Crenessity from 0% to 23%.

Cross-domain (paragraph corpora, coverage judge on reference facts; previous round r6 in brackets): FDA 67.0% (49.2%), statistics bulletins 55.5% (54.7%), NVDA 38.3% (56.7%). Not stated 1.6% / 0% / 0%; reads alone = false 7.4% / 5.0% / 0%. The NVDA drop is entirely two table fragments without column headings (#28, #37): the previous round covered them with statements whose phrase was the number itself, which the contract forbids ("leave out cells whose column heading is not in the passage"); the prose units are identical between the two wordings and better than r6 (54/73 against 41/68). The first wording of this change, which said only that a generic phrase in an adjunct is not a thing, made the model drop "patients with decompensated cirrhosis" as an object and raised FDA misworded to 9.7%; the sentence "a described thing that is the subject or the object is listed like any other" brought it back to 5.7%.

The sentence that started this, "各地区要深入社区、企业等基层一线开展形式多样的宣传活动", used to give 深入→社区 and 深入→企业; it now gives 各地区 —要深入→ 社区、企业等基层一线 with {开展: 形式多样的宣传活动}, and neither 社区 nor 企业 is a node.

The NVDA earnings release (52 chunks, 397 statements under #741) was still extracting when this was opened; its numbers follow in a comment.

## Not in this change

The modality qualifier and the implicit subject of directive sentences (prior-work item 2); table rows carrying their section (the period-column follow-up of #729); the unfolded description of rule 4 as a candidate for the same "reads alone" treatment; the second-look pass for figures (item 5); the quote before the statement in the contract (item 9).


WaylandYang:
The NVDA earnings release (52 chunks) finished under this contract. Old contract against this one, judged by the same second model:

| earnings release | #741 | this change |
|---|---|---|
| statements | 396 | 295 |
| not stated | 1.0% | 0.0% |
| misworded | 10.1% | 19.0% |
| reads alone = false | 15.2% | 21.7% |

Whole batch (25 documents, 846 statements): stated 88.7%, misworded 10.6%, not stated 0.7%, reads alone = false 15.6%.

The regression is the tables, and it is one mechanism: 38 of the 56 misworded and 43 of the 64 unreadable statements in this document come from table rows, almost all of the shape "Basic —three months ended→ $1.08", where the period column heading became the phrase and the row label lost its parent line (earnings per share). The chunks these come from carry a fake header row (`| NVIDIA CORPORATION | | |` followed by the separator) and none of the real column headings, so the model has no way to do better; the contract's "phrase is the column heading" then puts the period where the measure should be.

That is the next slice, in the reader rather than the contract: a table is linearized by code, one row to one line with the caption, the header rows and the row's parent labels repeated in every chunk; the contract then says that a heading that names a period is `when`, a unit is a qualifier, and the thing the caption names is the subject. Two structural drop reasons follow from what this batch shows (phrase equal to the value, phrase equal to the subject). Prose is unaffected: the 24 short documents and the three paragraph corpora stand as in the description.

PR #756, opened 2026-09-18
Give the ontology table view a way to create classes and properties

zhongqishang:
## Problem

The ontology page loses its only create affordance the moment you switch to the table view. `New class` and `New property` sit on the first row of the left rail, and that whole block is folded away in the table view (`.u-rail-fold.is-folded`, `opacity: 0` plus `pointer-events: none`) — reasonably so, since the table is itself the list.

What was left had no way to start from nothing. The only routes in were through an already-selected class's panel: new sub-class, a relationship with that class pre-filled as its domain, an attribute. A top-level class, or a relation that declares no domain at all, could not be created without switching back to the graph view.

## Change

A primary button at the right end of the table toolbar, following the active tab: `New class` on Classes, `New property` on Properties. It sets the same `setEdit` state as the rail row, so what opens is identical — no parent class, no initial domain, both are picked in the dialog. Disabled while the ontology is still loading, because a new class needs a parent list that is empty at that point.

The Attributes tab deliberately gets no button: an attribute always belongs to a class, and `AttributeDialog` takes its domain from whoever opens it instead of asking, so that entry point stays where it is, in a class's panel.

## Measured

`pnpm build` succeeds (5.93s) and `npx tsc --noEmit` is clean.

## Tests

None added. This wires an existing dialog into one more entry point; there is no new logic to test.


WaylandYang:
Merged. I checked each claim rather than taking it on trust: the state the new button sets is byte-identical to the rail row's, so the same dialog opens with no parent and no domain preselected; on Classes the dialog shows "Parent class (top level)" empty; Properties gives its own button; Attributes correctly gives none, and your reason for that is the right one.

Thank you for finding this. The rail folding away in table view took the only path to a top-level class with it, and nothing in the table said so.

PR #761, opened 2026-09-18
Fix GBK-encoded PDF text extraction

### LINKED ISSUE #739: Upload document unsupported encoding GBK-EUC-H
thread 'tokio-rt-worker' (190) panicked at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/pdf-extract-0.12.0/src/lib.rs:986:21:

unsupported encoding GBK-EUC-H

2026-09-17T01:08:53.352381Z  WARN utopia_store::jobs: 任务执行失败 job_id=37 kind=process_document error=task 38779 panicked with message "unsupported encoding GBK-EUC-H"

Floating-Y:
## Summary
- Catch failures from the current PDF text extractor and fall back to Poppler's `pdftotext`, including for the reported `GBK-EUC-H` panic.
- Preserve the existing OCR path when a PDF has no extractable text.
- Install Poppler and CJK CMap data in Docker and CI, and document the source-install requirement.

## Tests
- Added a synthetic `GBK-EUC-H` PDF regression test containing only test text; no user document or sensitive sample is included.
- `cargo test -p utopia-ingest` (91 passed)
- `cargo fmt --all --check`
- `cargo test --workspace` and `cargo clippy --workspace --all-targets -- -D warnings` passed before rebasing onto the latest `dev`.

Docker image build was not run locally because the Docker engine was unavailable.

Fixes #739


WaylandYang (commented):
Thank you for this — the diagnosis is right and the regression test is the part I appreciate most: a synthetic PDF with a named CMap and no `ToUnicode` map, so nobody needs the reporter's document to see the bug again.

I checked two things before reviewing the code.

**The crate version is not the fix.** We pin `pdf-extract = "0.12"` and the report came from 0.12.0. On 0.12.1 the same synthetic PDF panics at the same line, `lib.rs:986`, with `unsupported encoding GBK-EUC-H`. So the fallback is doing real work.

**Both new tests ran in CI and passed**, on the runner where the workflow now installs `poppler-utils poppler-data`.

## One change I would like before this goes in

On my machine `gbk_euc_h_pdf_extracts_chinese_text` fails, and the way it fails is the thing to fix. Git for Windows ships an executable called `pdftotext` that is **Xpdf 4.00, not Poppler**. It exits 0 and prints nothing for this PDF, so `pdf_with_poppler` returns `Ok("")`, `pdf()` returns an empty string, and `parse` reports:

```
This scanned PDF has no text layer; reading it needs a document-reading (OCR) service
```

The document has a text layer. We could not read it. Telling the operator it is a scan sends them to configure an OCR service for a file that OCR was never needed for, and the panic that explains everything is gone by then.

The empty-text case is only allowed to mean "no text layer" when the extractor itself said so. When the first extractor panicked or errored, an empty fallback result is a reader failure and should carry the original message:

```rust
match pdf_with_poppler(bytes) {
    Ok(text) if !text.trim().is_empty() => Ok(text),
    // The first extractor said there is nothing to read: the OCR path takes it from here.
    Ok(_) | Err(_) if original_error.is_none() => Ok(String::new()),
    Ok(_) => anyhow::bail!(
        "PDF text-layer extraction failed: {}; the fallback returned no text (is this Poppler, with its CJK CMap data?)",
        original_error.unwrap_or_default()
    ),
    Err(error) => anyhow::bail!(
        "PDF text-layer extraction failed: {}; Poppler fallback failed: {error:#}",
        original_error.unwrap_or_default()
    ),
}
```

Same shape covers the other half of it: Poppler without `poppler-data` reads this file as nothing at all, and an operator who installed only `poppler-utils` would get the same misleading message. The error text is where that hint belongs.

## Two smaller notes

- `catch_unwind` leaves the panic message on stderr, so every GBK PDF logs a Rust panic even though we handled it. Not blocking.
- Spawning a separate process for a file we already know one parser choked on is a quiet safety gain, since a crash in the child cannot take a worker down with it. Worth a line in the code comment, because it reads like a workaround and it is also a containment.

Everything else looks right to me, including keeping `Ok(String::new())` so a real scan still reaches the OCR path, and documenting the source-install requirement in both CONTRIBUTING files.

WaylandYang:
I pushed the change to your branch rather than asking you to do it, since I had the reproduction in front of me. Two commits are now yours plus one of mine; please read it and say if you disagree with any of it.

**What I changed and why my first suggestion was wrong.** I asked for "an empty fallback after a panic is a reader failure". That breaks `empty_gbk_euc_h_pdf_still_requests_ocr`: a PDF that declares a GBK font panics the extractor whether or not it draws anything, so after a panic an empty fallback is ambiguous. A scan and a file we cannot read both come back as empty text and exit code 0.

The file itself settles it. `draws_text` loads the PDF with `lopdf` and asks whether any text-showing operator (`Tj`, `TJ`, `'`, `"`) carries a non-empty string:

- draws nothing: a scan, `Ok(String::new())`, the OCR path takes it, exactly as before;
- draws glyphs and nothing came back: we could not read it, so the error carries the original panic and names the likely cause.

`lopdf` moved from dev-dependencies to dependencies. It costs nothing: `pdf-extract` already depends on it, so it was in the tree.

**The test now skips unless `pdftotext` is Poppler,** and `UTOPIA_TEST_REQUIRE_PDFTOTEXT=1` in the backend job stops that skip from being silent in CI, the way `UTOPIA_TEST_REQUIRE_DB` already does for the database tests. This matters because the two programs called `pdftotext` are not the same program: Git for Windows ships Xpdf 4.00, which returns nothing for your fixture, and before this change that turned into "this scanned PDF has no text layer" on my machine.

Checked: `cargo test -p utopia-ingest` 91 pass here with the fallback test skipping, the same test fails with the requirement set (which is what CI will enforce), workspace fmt and clippy clean.

One thing worth doing separately, if you are interested: `pdf-extract` panics rather than returning an error for an unknown CMap, and 0.12.1 still does. An upstream patch that maps the predefined CJK CMaps, or at least returns `Err`, would let this fallback become a rare path instead of the one every CJK PDF takes.

PR #763, opened 2026-09-19
Keep late chat callbacks from finishing a follow-up answer

Maya-Kid:
## Problem

A follow-up answer can lose its streaming state when the previous request finishes closing. `consumeChatStream` calls `onDone` on the `done` event and again at EOF. Between those callbacks, the user can send another message in the same conversation. The old handle's `finish()` then looks up the reused conversation key and finishes the new answer, hiding its stop control and allowing another submission.

## Change

Bind each live-answer handle to its original slot identity, in addition to its conversation key. Late `finish`, `patchLast`, `identify`, and `setAbort` calls become no-ops once that slot has been replaced. Pending identification and independent conversations retain their existing behavior.

This completes the ownership protection introduced in #259 for successive generations within one conversation. No dependency, schema, public API, or UI text changes.

## Validation

- Added a regression using the real `streamChat` and `liveAnswer` with a controlled Fetch response: deliver `done`, start a follow-up, then close the old stream. The original code incorrectly sets the follow-up's `streaming` to false.
- Added stale-handle coverage and a control for pending IDs / independent conversations. Before the fix: 2 failures, 1 pass. After: all 3 pass.
- `pnpm@10.2.1 test`: 66 tests passed.
- `pnpm@10.2.1 build`: passed (style guard, TypeScript, Vite).
- Rust 1.98.1: workspace fmt, clippy with `-D warnings`, build, and non-DB workspace tests passed (DB tests early-return in that run).
- Fresh isolated PostgreSQL 16 / pgvector migration and migration-runner rerun passed.

DB-enabled server tests: 292 passed, 1 ignored, and the existing wall-clock assertion in `the_embedding_gate_is_never_held_beyond_its_ceiling` failed. It also fails in an isolated rerun: 1.913s vs the 0.9s assertion. This run accesses PostgreSQL over a LAN SSH tunnel, so latency is a possible contributor; it has not been established as the cause. No backend code or test threshold was changed. The non-DB baseline skips this DB-dependent test.

The full DB-required store run completed: 284 passed, 0 failed, 0 ignored (`UTOPIA_TEST_REQUIRE_DB=1`). These use a disposable remote database, not an existing deployment. Live-model calls and browser E2E were not run.


WaylandYang:
Verified here, and the part I checked hardest is that the regression fails without the fix: with `web/src/liveAnswer.ts` reverted to `dev` and the rest of your branch in place, two of the three new tests fail, including the one that drives the real `streamChat` and closes the old stream after the follow-up begins. With your change, 66 tests pass.

Binding the handle to its slot object rather than to the conversation key is the right shape. The key is not an identity: `identify` rewrites it mid-stream, and a follow-up reuses it, so every one of those four callbacks was reachable by a handle that no longer owns anything. Comparing `lives.get(key) === slot` says what the code means, which is "am I still the answer in this slot".

Two things I confirmed while reading, so they are on the record rather than in anyone's head:

- `identify` keeps working after it moves the slot, because `owned()` closes over the mutable `key` and the slot is re-inserted under the new one. Your third test covers exactly that, along with two pending answers in different bases staying independent.
- `stop` still reaches the current abort and never the stale one, which is what the second test asserts.

On the `the_embedding_gate_is_never_held_beyond_its_ceiling` failure you reported: that assertion is wall-clock, and it also fails on the first run against a cold local database here and passes on a rerun. Over an SSH tunnel it is expected to fail. Nothing for you to do about it in this pull request, and thank you for reporting it instead of quietly changing the threshold.

PR #766, opened 2026-09-19
Reject impossible calendar dates before converting clock inputs

Maya-Kid:
## Problem

Date correction and review forms accept impossible calendar dates when a clock time is supplied. For example, `2023-02-29T10Z` is silently normalized by JavaScript to March 1 and sent to the API as a valid timestamp. The same invalid day without a time is already rejected.

`parseDateInput` validates calendar round-tripping only in its date-only branch. The timestamp branch checks only whether `Date` is NaN, which does not detect this normalization. Its callers include the fact-time editor and review decision forms.

## Change

Run the existing calendar-date validation for both branches, before timezone conversion. Valid offsets that cross a day or year boundary remain accepted. No UI copy, dependencies, API contracts, or backend behavior changes.

## Validation

- Original code: 5 new invalid-timestamp cases fail; 11 controls pass.
- Fixed code: all 16 cases pass, covering non-leap February 29, February 30, April 31, positive/negative offsets, valid leap days, cross-day/year conversion, date-only precision, and missing timezones.
- Full frontend suite: 79 passed.
- `pnpm@10.2.1 build`: passed (style guard, TypeScript, Vite).
- `git diff --check`: passed.

No database writes or browser E2E were performed for this frontend-only correction. The isolated tests exercise the actual shared parser used by the forms.


WaylandYang (commented):
Clean, and the one I would take as it stands.

Verified here. On the branch, 16 tests in `time.test.ts` pass and the whole web suite is 79 across 13
files. With only `web/src/time.ts` reverted to `dev` and your test kept:

```
Tests  5 failed | 11 passed (16)
AssertionError: expected { …(2) } to be null
+ Received: { "iso": "2023-02-28T16:30:00.000Z", "precision": "minute" }
```

Exactly the five new cases, which is what a regression test is for. Hoisting the round-trip check
above the `hh === undefined` split is the right shape: the calendar date is written by a person and
the offset is arithmetic applied afterwards, so a legitimate offset-driven day shift still works and
February 29 in a non-leap year does not.

Two notes, neither blocking:

- `T24:00` still rolls into the next day (`2024-02-29T24:00Z` gives `2024-03-01T00:00:00.000Z`).
  That is ISO-legal and unchanged by this pull request, but a form that refuses February 30 while
  quietly accepting hour 24 is uneven. Worth a case one day, not this one. The genuinely
  out-of-range values are handled: `25:00`, `10:60` and `10:15:60` all return null.
- The comment you replaced was Chinese (`// 溢出静默：2023-13-01 会变成 2024-01-01…`) and the new one
  is English. Comments in this repo are Chinese; English is for records, pull requests and commit
  messages. Same note applies to your other four.

PR #767, opened 2026-09-19
Harden embedding response index handling

Maya-Kid:
## Scope and evidence

This is defensive hardening, not a confirmed bug affecting a supported provider. A local HTTP fixture shows that the client associates vectors with the wrong input texts if an endpoint returns embedding entries out of index order. I have not identified a supported provider that produces such responses, and the presence of `index` in the [response schema](https://developers.openai.com/api/reference/resources/embeddings/methods/create) alone does not establish that reordered arrays are permitted.

## Change and tradeoff

When indices are present, assemble vectors by index and reject duplicate, missing, invalid, or out-of-range indices. Preserve positional behavior when the entire response omits indices, matching existing compatible-endpoint fixtures. The stricter validation may reject endpoints whose index metadata was previously ignored; no live-provider compatibility matrix has been tested. No dependencies or public API changes.

## Validation

- Local HTTP tests reproduce reordered vectors and accepted duplicate indices before the change; the unindexed compatibility control passes.
- After the change, all 26 `utopia-llm` tests pass, including invalid-index coverage.
- Workspace fmt, clippy (`--all-targets -- -D warnings`, Rust 1.98.1), and `git diff --check` pass.

No live embedding provider or database was called. This patch does not establish or repair existing data corruption.


WaylandYang (commented):
The hazard is real: `embed` zips response order against input order, and `pipeline::embed_pending` only checks the count, so a reordered batch stores a vector against the wrong text and nothing anywhere would notice. Scattering by `index` is the right fix, and keeping the positional path for responses that carry no index is the right compromise.

Verified. With the fix, 3 tests pass and the crate is 26 green. With the hunk reverted and the tests kept:

```
---- tests::indexed_embeddings_follow_the_input_order stdout ----
assertion `left == right` failed
  left: [[2.0, 20.0], [1.0, 10.0]]
 right: [[1.0, 10.0], [2.0, 20.0]]

---- tests::ambiguous_embedding_indices_are_rejected stdout ----
accepted [{"embedding":[1.0],"index":0},{"embedding":[2.0],"index":0}]
```

## One line I would like changed before this goes in

```rust
data.iter().any(|item| item.get("index").is_some())
```

`get` returns `Some(Value::Null)` for `"index": null`, so a response that spells the field and leaves it empty takes the indexed path, `as_u64()` fails, and the whole call errors. Same for a string index or a float one. All three work today:

```
PROBE explicit nulls: Err(Embedding response has a missing or invalid index)
PROBE string index:   Err(Embedding response has a missing or invalid index)
PROBE float index:    Err(Embedding response has a missing or invalid index)
```

Your own comment says the goal is to keep the positional format for endpoints that do not index, and a gateway that writes `"index": null` is exactly such an endpoint. The predicate should test the value, not the key:

```rust
data.iter().any(|item| item.get("index").and_then(serde_json::Value::as_u64).is_some())
```

A response that carries a real index on some entries and a null on others then falls to the positional path, which is no worse than today and does not turn a working endpoint into a hard failure.

## Two notes

- The title says "paired with their input texts", and the no-index path still mispairs on a length mismatch (`fewer entries, no index` returns one vector for two texts). It is not exploitable: every caller already bails on a mismatch, in `pipeline.rs:290`, `ontology_index.rs:112` and `:170`, `mapping_index.rs:66`, and `bootstrap_ontology.rs:151`. So the ordering half is the real gap and you closed it. Worth a sentence saying the length half lives at the call sites.
- `embeddings_without_indices_keep_positional_compatibility` passes with the source reverted. That is correct for a compatibility guard, and it is worth knowing it proves nothing about the fix.

New error strings are English, ten lines from a Chinese one in the same function. Comments and messages in this repo are Chinese.

WaylandYang:
Landed as #784, with your commit and its authorship. The branch could not be rebased from here, so it moved rather than being merged in place. One change on top, from the review: the index-presence probe now reads the value rather than the key, because `get("index")` returns `Some(Value::Null)` and a null, string or float index turned three working response shapes into hard errors. Thank you for catching the ordering hazard.

PR #768, opened 2026-09-19
Reject model streams that end before completing their turn

Maya-Kid:
## Problem

A clean HTTP response ending before the model sends `[DONE]` or `finish_reason` is treated as success by `chat_stream_raw` and `chat_tools_stream_with`. The tool path can emit a completed turn containing partial JSON arguments. The collected `chat_at_streaming` path already detects this condition using the typed `Interrupted` error.

## Change and user-visible behavior

Recognize either finish signal in both live readers. If EOF arrives without one, emit `Interrupted` and do not emit an unfinished tool turn. Deltas received before the interruption remain streamable.

**In the chat path that consumes `chat_stream_raw`, the user sees the partial text followed by an error, and the partial assistant turn is not saved in conversation history.** The error returns before `conversations::append_message`; previously that shortened answer was saved with a clean `done`. This intentionally keeps an incomplete answer from being recorded as a completed turn.

All three readers now count Unicode characters in `Interrupted.got`, matching the error message, rather than counting UTF-8 bytes. No dependencies or existing public API changes. Empty live streams retain the existing `Interrupted { got: 0 }` diagnosis; this patch does not unify that with the collected reader's distinct no-frames error.

## Validation

- Real local HTTP fixtures reproduce successful partial answers and unfinished tool arguments before the fix; both become `Interrupted` afterward.
- Controls cover completion by either `[DONE]` or `finish_reason`.
- New character-count regression: `你好🦀` was reported as 10 characters; all three readers now report 3.
- Full `utopia-llm` suite: 27 passed. Workspace fmt, clippy (`--all-targets -- -D warnings`, Rust 1.98.1), and `git diff --check` pass.

No live model or database was used. The branch now includes the merged UTF-8 buffering fix from #764 via dev. Endpoints that omit both completion signals are rejected, as the collected reader already does.

After merging dev (`5cf3eac`) and resolving the shared raw-stream initialization, all 30 LLM tests pass, including the Unicode buffering and interruption regressions together. LLM clippy (`--all-targets -- -D warnings`), workspace fmt and diff checks pass.


WaylandYang (commented):
The gap is real and it is mine: #757 gave `chat_at_streaming` the completion check and left the two live-stream paths without it. Carrying the same `Interrupted` through them is right.

Verified. With the fix, 3 new tests pass and the crate is 26 green. With the five hunks reverted and the tests kept:

```
---- tests::a_tool_stream_cut_before_its_finish_is_an_error stdout ----
called `Result::unwrap_err()` on an `Ok` value: [Turn(AssistantTurn { content: None,
  tool_calls: [ToolCall { id: "call_1", name: "lookup", arguments: "{\"name\":" }] })]
```

That line is the whole argument for this change. Without it the agent is handed a tool call whose arguments are `{"name":` and has to discover for itself that the JSON is a fragment.

## One thing the description should say out loud

`api/chat.rs:927` streams `chat_stream_raw`, and on `Err` it yields an error event and returns **before** `conversations::append_message`. So a cut stream now shows the reader the partial text followed by an error, and the assistant turn is not saved at all. Before this change it was saved, short, with a clean `done`.

I think losing a half answer is the right call, and it matches what the ledger does everywhere else: half a thing is not a thing. But it is a product decision riding inside a correctness fix, and the pull request body does not mention it. Please add a line, so that whoever reads the record later sees it was chosen rather than inherited.

## Two smaller

- `chat_at_streaming` separates "no frames at all" (`LLM stream carried no frames`) from "cut mid-answer" (`Interrupted`). The two new checks collapse both into `Interrupted { got: 0 }`, so one file now describes the same endpoint failure two ways. Not worth its own change, worth knowing.
- `got` is bytes while `Interrupted`'s message says chars. Pre-existing, inherited from the line I wrote, now in two more places. If you are touching it anyway, `chars().count()` costs nothing at that size.

Not a concern, an observation: an endpoint that sends neither `finish_reason` nor `[DONE]` now fails every call and burns the retry budget doing it, since `transient()` treats `Interrupted` as retryable. #757 set that precedent for one path and this extends it to three. The in-repo fake server always sends `[DONE]`, so nothing here breaks. If we ever meet a shim that does neither, the fix is to say so in the endpoint's settings rather than to soften the check.

WaylandYang:
Landed as #785, with your commit and its authorship; the branch could not be rebased from here after #764 moved the streaming buffer to bytes, so it moved rather than merging in place. The conflict was mechanical: keep the byte buffer, add your `ended` and `got`. The pull request body now states the consequence in `api/chat.rs` — a cut stream no longer persists the partial assistant turn — so it reads as chosen rather than inherited. Thank you.

PR #776, opened 2026-09-19
The token ceiling is ours, and a cut reply says so (#760)

### LINKED ISSUE #760: The endpoint decides where a long answer stops
Extraction never sends `max_tokens`, so where a long answer stops is decided by the endpoint's own
default. Measured while trying a reasoning cap (#759):

- With reasoning disabled, a dense passage's answer stopped at exactly 4,096 completion tokens with
  `finish_reason: length`, and the JSON could not be parsed.
- The four-filing run with `reasoning_effort: low` recorded 8 `truncated_reply` drops; the uncapped
  baseline recorded 0. Across all bases in the e2e database the reason has fired 10 times.

`truncated_reply` already does the right thing: the parser keeps every complete item and records a
drop row, so nothing is silently lost. Two things are still worth doing:

1. **Send an explicit `max_tokens`** on the extraction and alignment calls, large enough that a
   dense chunk's JSON fits, so the ceiling is ours rather than whatever the endpoint defaults to.
   This endpoint ignores the field when reasoning is on, and honours it when reasoning is off, so
   sending it costs nothing and helps wherever it is read.
2. **Say why the answer was cut.** `chat_at_streaming` treats `finish_reason: length` as a normal
   end, so a truncated answer reaches the parser looking complete. Carrying the reason through would
   let the drop row say the answer hit the token ceiling instead of only that it did not parse.

Neither is urgent: the drop is a row today and the statements before the cut are kept.

basil-k-aji-dev:
Fixes #760.

## What it does

**The ceiling is ours.** The streamed calls — extraction, both aligners and time resolution, all of which go through `chat_at_streaming` — now send an explicit `max_tokens` of 16,384. Without it the ceiling was whatever the endpoint defaulted to, which is how a dense passage's answer stopped at exactly 4,096 completion tokens with `finish_reason: length` and left JSON that could not be parsed.

16,384 is four times the longest answer measured on this path: `docs/design/extraction.md` records about 1,500 completion tokens for a dense chunk's answer, and 4,096 is the line actually hit.

**A cut reply says so.** `take_frame` saw `finish_reason` and set `ended = true`, discarding which reason it was. Both `stop` and `length` end the stream, but a `length` reply is half an answer that looks exactly like a whole one, so by the time the parser failed the only thing left to record was that it did not parse. `chat_at_streaming` now returns a `Reply { text, finish_reason }`, and `extraction_open` uses it: a chunk that hits the ceiling gets a drop row saying so, rather than one that says only that the reply was unparseable. An answer that is too long is a number to change; a reply that does not fit the shape is not, and the drop table can now tell a reader which one happened.

## What it does not do

Sending the field is not a guarantee, and the PR should not be read as one. The measurement in #759 stands: this endpoint **ignores `max_tokens` with reasoning on**, where a call capped at 4,096 returned 45,428 completion tokens and finished normally. It is honoured with reasoning off, which is the configuration the truncation was observed in. The field costs nothing where it is not read, which is the argument for sending it rather than a claim that it fixes truncation everywhere.

`finish_reason` is `Option<String>`, not a bool: some implementations only send `[DONE]`, and its absence does not mean the answer was complete.

## Changes

| File | What |
|---|---|
| `crates/utopia-llm/src/lib.rs` | `MAX_COMPLETION_TOKENS`, `Reply { text, finish_reason }` with `hit_token_ceiling()`, `take_frame` records the reason |
| `crates/utopia-server/src/extraction.rs` | `chat_retrying_rate_limits_at` returns `Reply` |
| `crates/utopia-server/src/extraction_open.rs` | drop rows distinguish the token ceiling from an unparseable reply |
| `type_alignment.rs`, `phrase_alignment.rs`, `time_resolution.rs` | `&reply` → `&reply.text` |
| `docs/design/extraction.md` | "Where a long answer stops", next to the #759 note it follows from |

## How this has been tested

`a_token_ceiling_is_ours_and_a_cut_reply_says_so` covers both halves, because they are two halves of one fault: without the field the endpoint picks the ceiling, and with the field but no reason carried through, a cut reply is still silent about why.

The existing test server discarded the request, so `read_request` now returns it and `an_http_response_capturing` hands it back — the assertion is on what we *send*, not only on what the endpoint answers.

Each half was checked to fail on its own. Removing the `max_tokens` line:

```
assertion `left == right` failed: 不送这个字段，答案断在哪里就是端点的默认值说了算：
{"messages":[...],"model":"m","stream":true,"stream_options":{"include_usage":true},"temperature":0.0}
  left: Null
 right: Number(16384)
```

Removing the line that records the reason:

```
thread '...a_token_ceiling_is_ours_and_a_cut_reply_says_so' panicked at crates/utopia-llm/src/lib.rs:1176:9:
None
```

Run locally on `aarch64-unknown-linux-gnu`, rustc 1.98.1:

- `cargo fmt --all --check` — clean
- `cargo build --workspace` — clean
- `cargo test --workspace`

The tests that need `UTOPIA_DATABASE_URL` skip here rather than run; I have no Postgres on this machine, so that part of the suite is untested by me and left to CI.


WaylandYang (commented):
Verified both halves here, each on its own, because that is the claim the test makes.

Removing the field:

```
assertion `left == right` failed: 不送这个字段，答案断在哪里就是端点的默认值说了算
  left: Null
 right: Number(16384)
```

Dropping the reason instead, by deleting the one line that records it:

```
thread 'tests::a_token_ceiling_is_ours_and_a_cut_reply_says_so' panicked at crates\utopia-llm\src\lib.rs:1189:9
```

24 `utopia-llm` tests pass on the branch, workspace clippy and fmt are clean. Asserting on the
request body rather than only on what the fixture answers is the right call: what we send is the
thing this pull request changes.

## One number to reconsider: 16,384

It is chosen against the answer — four times the ~1,500 tokens a dense chunk's JSON costs — and on
this endpoint today that is safe, because the field is ignored with reasoning on. The field is now
sent on every streamed call, though, and `max_tokens` does not mean the same thing everywhere.
Anthropic counts thinking inside `max_tokens`; OpenAI's `max_completion_tokens` counts reasoning
tokens. On an endpoint that counts reasoning, 16,384 is not a ceiling on the answer, it is a budget
for the thinking.

That is the lever #759 measured and rejected. `reasoning_effort: low` held this model to 15.5k and
16.5k reasoning tokens, right about 16,384, and cost 885 statements against 729, with table figures
reaching no statement going from 2% to 16%. A cap that lands there by accident would do the same
damage silently, and the drop row would say the reply hit the token ceiling, which would be true and
would read like a bug in the answer rather than a budget we set.

The guard you want is against the endpoint's 4,096 default, and any number well above a real
completion does that. The largest completion measured on this path is 45,428 tokens (#759, reasoning
on). **65,536 keeps the guard and can never become the reasoning cap.** Alternatively send the field
only where reasoning is known to be off, but that is state we do not have, so I would just raise the
number.

Either way the paragraph in [extraction.md](docs/design/extraction.md) should say which quantity the
number is measured against, because a reader who sees "four times the longest answer" will not
expect it to bound the thinking on another endpoint.

## Smaller

- `finish_reason: Option<String>` rather than a bool is right, and the comment saying that its
  absence is not a completeness claim is the part someone would otherwise have to rediscover.
- The non-streamed `chat_at` and `chat_tools` do not get the field. That matches #760, which asked
  for the extraction and alignment calls, and I would leave them alone.
- The drop row now distinguishes the two causes, which is what the issue was for. Nice that the
  detail line keeps the chunk number and the parser's own error.

WaylandYang:
Landed as #786, with your commit and its authorship; the branch could not be rebased from here after #764 moved the streaming buffer to bytes. One change from the review: 16,384 became 65,536, because the field now rides every streamed call and `max_tokens` counts reasoning on some endpoints — at 16k it would silently become the reasoning cap #759 measured and rejected. The largest completion measured on this path is 45,428, so the new number can only ever guard against the endpoint's own default. The constant's comment and the design page now say which quantity it is measured against. Thank you for a clean piece of work, and for writing down what it does not promise.

basil-k-aji-dev:
Thank you — and the change to 65,536 is a better reading of the measurement than mine was.

I set 16,384 against the answer: ~1,500 tokens for a dense chunk's JSON, times four. That silently assumes `max_tokens` bounds the answer, and on an endpoint where it counts reasoning it does not — it becomes a reasoning cap, which is exactly the lever #759 measured and left alone. Picking a number four times the thing I was not bounding would have re-introduced the rejected change through a side door, and the e2e numbers would have moved without anyone having asked for it.

Measuring against the largest completion seen on the path, 45,428, is the right quantity, and it keeps the field doing only the one job it can do everywhere: guard against the endpoint's own default. Putting which quantity it is measured against into the constant's comment and the design page is worth more than the number itself — that is the part that stops the next person picking 16,384 for the same reason I did.

No objection to landing it as #786; rebasing across #764 was not going to be clean from here.

PR #778, opened 2026-09-19
Keep XML references in imported Office text

Maya-Kid:
Importing a DOCX or PPTX containing `R&D` currently produces `RD`; `2 < 3` loses its comparison sign. The XML reader emits character references separately from `Text`, and both Office text readers discard those events. CDATA text is also discarded.

Handle references and CDATA inside text runs, including Word table cells. Known references are decoded once; unknown references are preserved literally, matching the WebDAV reader, so an unrecognized entity does not abort the whole import. CDATA remains literal. This changes no dependencies or public APIs.

Validation against `dev` at `e26d97aaed5f75472d6bab7dfa01a0796dcd9c75`:

- Three regressions call the public `utopia_ingest::parse` entry point with minimal ZIP fixtures containing the relevant Office XML parts. All three fail before and pass after the fix on both macOS and Linux (Rust 1.98.1).
- macOS: all 98 `utopia-ingest` tests pass, including the PDF integration tests.
- Linux: all 93 unit tests plus the three new integration tests pass in an isolated Rust container.
- Workspace fmt, clippy (`--all-targets -- -D warnings`), and `git diff --check` pass.

No database or model calls are involved. Already imported documents need reprocessing to recover lost characters. Related to the XML event handling in #765, but this patch only changes the independent DOCX/PPTX ingestion readers.

Review follow-up: a fourth integration test covers an unknown `&nbsp;` and an internal-DTD entity in both Office readers, alongside a known `&amp;`. It fails before the compatibility fix and passes afterward. Latest macOS ingest run: 99 tests passed (93 unit, 4 Office integration, 2 PDF integration); ingest clippy, fmt and diff checks pass. The CDATA test was renamed to describe its regression; the existing out-of-run guard remains a control.


WaylandYang (commented):
Right diagnosis, right places, and the tests are real: reverting `parsers.rs` alone and keeping your tests gives

```
test result: FAILED. 0 passed; 3 failed
---- a_docx_keeps_references_in_paragraphs_and_table_cells stdout ----
RD: 2  3  1; quote word; ; lt;
| Team | Budget |
| RD | 100 |
```

This also completes the sweep. Three quick-xml reader loops exist in the repository — the two you fix here and `webdav.rs`, which had the same bug and was fixed in #765. Nothing else in the tree reads XML this way: calamine handles `GeneralRef` itself, oxrdfxml pins a quick-xml old enough that entities still arrive inside `Text`, and the HTML path goes through html5ever.

## One line to change: `?` turns a lost character into a lost document

```rust
let s = quick_xml::escape::unescape(&reference)?;
```

`unescape` only knows the five predefined entities, so anything else — `&nbsp;` from a non-Microsoft producer, an entity declared in an internal DTD — now fails the whole import:

| | on `dev` | on this branch |
| --- | --- | --- |
| docx with `Paris&nbsp;2024 and R&amp;D` | `"Paris2024 and RD\n"` | `at 1..5: unrecognized entity 'nbsp'` |
| pptx, same content | `"\n## Slide 1\nParis2024 and RD\n"` | `at 1..5: unrecognized entity 'nbsp'` |

Before your change the reader dropped one character. After it, the file does not import at all. That is a strictly worse failure for the case your change does not cover.

#765 hit this exact question in `webdav.rs` and answered it the other way, and that answer is now on `dev`:

```rust
match quick_xml::escape::unescape(&reference) {
    Ok(v) => text.push_str(&v),
    Err(_) => text.push_str(&reference),
}
```

Please match it in both readers. Strict OOXML only uses the five predefined entities, so this needs an unusual producer to bite — which is exactly the kind of file where losing the whole document is worst.

## Two notes

- `office_cdata_is_literal_and_text_outside_runs_is_ignored` — the second half of that name passes identically with the fix reverted, because the existing `in_text` guard already handled it. Only the CDATA half is a regression test. The name promises more than it checks.
- Attribute values are still raw (`a.value.to_string()`). No impact today, since every caller parses a number out of `w:val`, `w:ind`, `w:gridSpan`. Worth knowing that the text path is now correct and the attribute path is not.

93 `utopia-ingest` tests pass on the branch, fmt and clippy clean.

PR #779, opened 2026-09-19
A citation opens before it leaves

WaylandYang:
Two things about citations in Chat.

**`[n]` in the answer was never clickable.** Not a streaming bug: nothing anywhere turned a marker into a control — `Segment` rendered the markdown as it came, and the only code that read `[n]` was `citedSources`, which decides what the footer lists. `rehypeCitations` now runs in the rehype pipeline, so it can see whether a bracket sits inside a link or a code span and leave those alone; a marker becomes `<a href="cite:n">` and `components.a` renders it as a chip. `[1, 2]` becomes two chips, because one click should not open two excerpts.

While an answer is streaming, a number whose source has not arrived yet renders as a plain muted `[n]` and becomes clickable when the source does — sources arrive one retrieval at a time and a dead control is worse than a quiet one. A half-typed `[2` stays out: the regex wants the closing bracket. `remend` repairs the same half-typed marker into `[2](streamdown:incomplete-link)`, which used to flash as a blue "2" mid-sentence; those are now rendered as their own text.

**A citation opens a preview instead of leaving the page.** Clicking a chip or a footer row opens a popover with the filename, where the text came from when it was not simply stated (0040), and the passage itself — the 160-character excerpt at once, replaced by the full chunk when the document detail arrives. "Open original" sits at the top right and goes where the row used to go directly. The preview reads through the same react-query key the document page uses, so following that link lands on a page that is already loaded.

Checking a sentence against its source usually needs the sentence, and leaving the page for it loses the answer's scroll position.

Verified: `tsc --noEmit` clean, style guard clean (86 files), `vitest run` 72 passing including 12 new ones over the marker shapes, the rehype transform and the opaque elements. The live look is not verified here — the preview harness serves `D:\projects\LandscapeBI\web` and this branch lives in a worktree, so the rendering is the one thing left to look at.


WaylandYang (commented):
The one thing this said was unverified — the live look — is the thing I could check, so I ran the branch on the preview at 5199 against a real conversation in the e2e base, one with seventeen markers in a Chinese answer over four NVDA filings.

It works:

- All seventeen `[n]` render as chips, as `button` elements in accent colour, including the `[1] [16]` pair that has to become two controls rather than one.
- Clicking one opens the preview anchored at the marker: the filename, **Open original** at the top right, and the passage itself.
- A click anywhere outside closes it.

Web checks on the branch: 72 tests across 13 files, `tsc --noEmit` clean, style guard clean, production build clean.

## One thing to look at

The excerpt renders the chunk's raw text, so a passage that begins with a markdown heading shows its marker. The first preview I opened reads:

```
nvda-q2fy27-earnings-release.html    Open original ↗
###
© 2026 NVIDIA Corporation. All rights reserved. NVIDIA, the NVIDIA logo, BlueField, …
```

That `###` is the chunk's own text, and the reader has no way to know it is not part of the quotation. Two defensible answers: render the excerpt as markdown, which risks a heading swallowing the popover's own type scale, or strip leading marker characters when showing a passage as a quote. I lean to the second, since this is a quotation of what the document says and the quote should read as prose.

## A note on what I could not test

I tried to check whether opening a second preview replaces the first. Clicking chips programmatically leaves two open, but programmatic clicks bypass the dismissal layer, and with real clicks I could not reproduce it — every real click outside closed what was open. So I am not reporting it as a defect, only saying I could not settle it. If you have a moment, click two markers in a row by hand and see what you get.

The rest is right: a marker whose source has not arrived renders plain and becomes a control when the source lands, and a half-typed `[2` staying out until its bracket arrives is the detail that would otherwise flash on every answer.

PR #780, opened 2026-09-19
Keep overlapping and superseded document reads from corrupting chunks

Maya-Kid:
Overlapping document processing can corrupt the live chunk set in two ways. Two replacements can both see no existing chunks and insert duplicate copies. A slow reader can also return after a source update has finished processing and replace the new text with old content, or return after deletion and recreate live chunks.

Lock the document row before reading the version and chunk claim pool. The pipeline additionally supplies the SHA of the input it read: while holding that lock, reject the replacement if the document is deleted, missing, or has a different SHA. An obsolete successful read then stops before indexing, updating ready status, or enqueueing extraction. Existing callers of `replace_chunks` retain their signature and behavior apart from serialized replacement. No schema or dependency changes.

The lock covers only database chunk replacement, not parsing or model calls. This protects reads superseded before replacement; it does not serialize the whole pipeline or cover source changes during later embedding/indexing stages or stale reader errors.

Validation:

- Real PostgreSQL overlap regression: before, two live rows for one input chunk; after, one live row and identical returned chunk IDs. Lock waits establish overlap without relying on sleeps.
- Two pipeline regressions use `process_document`, a paused local HTTP transcription fixture, real source upsert/deletion, and PostgreSQL. Before the guard, the old read overwrites a completed newer revision or recreates chunks after deletion. Afterward, the newer text and status stay intact, and deletion leaves no live chunks. The newer revision's processing job is already complete, so there is no pending retry that would repair the corruption.
- Linux tests on the branch including dev `87df718`: `utopia-store` and `utopia-server` report **591 passed, 0 failed, 1 ignored**, with `UTOPIA_TEST_REQUIRE_DB=1` against an isolated PostgreSQL/pgvector database. Both new regressions and the overlapping replacement regression pass. The ignored test requires public HTTPS; optional live-service tests have no provider configuration, so this is not a live-provider acceptance run.
- Workspace fmt, workspace clippy (`--all-targets -- -D warnings`, Rust 1.98.1), and diff checks pass.

No production data or real model endpoint was used.

The remaining workspace crates also passed in a separate Linux run (`cargo test --locked --workspace --exclude utopia-store --exclude utopia-server`). Optional external services and the PDF fallback tool were not configured.


WaylandYang (commented):
The race is real, the fix is in the right place, and the test is the best-built one I have read in this repository: it gates both writers on an actual database lock and polls `pg_stat_activity` instead of guessing at a sleep, and the same gate works for the fixed and the unfixed code, which is what makes it a discriminator rather than a coincidence. Reverting `documents.rs` alone:

```
assertion `left == right` failed: overlapping processing must not duplicate the document's live text
  left: 2
 right: 1
```

Deterministic across repeats, and green across repeats with the fix.

**It is also better than your description claims: it removes a deadlock that exists on `dev` today.** `restore_tx` and `purge` take the documents row and then touch chunks; `replace_chunks` on `dev` locks chunk rows first and only reaches the documents row at the insert's foreign-key check. That is a cycle, and it reproduces:

```
without this change (current dev) => A: insert ok | B: deadlock detected
with this change                  => A: insert ok | B: ok
```

Worth saying in the body. It is the stronger half of the argument.

## The one change I want: `FOR NO KEY UPDATE`

`FOR UPDATE` conflicts with the `FOR KEY SHARE` that every foreign-key check takes, so for as long as the transaction runs, nothing can insert a child row of that document. Measured by holding the lock and running each statement under a one-second `statement_timeout`:

```
[BLOCKED ] memory::append INSERT chunks (same doc)
[BLOCKED ] INSERT document_versions (same doc)
[BLOCKED ] set_status UPDATE documents (same doc)
[PROCEEDS] INSERT chunks for a different document
[PROCEEDS] UPDATE a different documents row
```

And the hold is not brief, because the lock is the first statement and the transaction is one round trip per chunk:

```
n=100  chunks: ~110 ms
n=1000 chunks: ~1.0 s
n=4000 chunks: ~4.1 s
```

So a long document blocks its own child-table inserts for seconds. `FOR NO KEY UPDATE` gives the same mutual exclusion against another `replace_chunks` without taking the foreign-key side:

```
holder=FOR UPDATE         [BLOCKED ] second replace_chunks   [BLOCKED ] INSERT chunks (FK)
holder=FOR NO KEY UPDATE  [BLOCKED ] second replace_chunks   [PROCEEDS] INSERT chunks (FK)
```

One word, and the blast radius becomes exactly what the title says.

## Three things the change does not do, worth a line each in the body

- **It fixes `replace_chunks`, not "document processing".** `set_status`, `reindex_document`, `embed_pending` and `set_ready` are still unsynchronized across two overlapping runs.
- **Last writer wins is untouched.** If the two runs carry different content, because the document was re-uploaded between enqueues, the lock serializes them and the one that commits second wins — which may be the one holding the older text. The guarantee is no duplicates, not the newest version.
- **Nothing is re-validated under the lock.** `.execute()` ignores `rows_affected`, so a document that does not exist takes no lock at all and the run proceeds; and if `delete()` commits while this transaction waits, it goes on to insert live chunks onto a soft-deleted document. Both predate this change, and `replace_content_and_enqueue_processing` shows the shape that closes them (`fetch_optional`, `NotFound`).

Minor: the two `max_connections(1)` pools in the test loop are never closed.

107 `utopia-store` test binaries pass on the branch. Two fail identically on `dev` with `relation "organizations" already exists`, which is the local database being already migrated, not this change.

WaylandYang:
Pushed the one-word change to your branch: the new lock is `FOR NO KEY UPDATE`. It gives the same mutual exclusion against another `replace_chunks` — your regression still passes, and still fails with the lock removed — without conflicting with the `FOR KEY SHARE` that every foreign-key check takes, which was blocking inserts into `chunks`, `document_versions` and `memory::append` for the whole transaction. That transaction is one round trip per chunk, so a four-thousand-chunk document was holding it for about four seconds.

The rest of what you added since the review is more than I asked for and the right more: rejecting a replacement whose SHA no longer matches closes the case where a slow reader returns and overwrites newer text, which I had listed as something the first cut did not do. 110 store test binaries pass here; the one failure is a self-migrating test against an already-migrated local database, which fails the same way on `dev`.

PR #712, opened 2026-09-15
Add utopia CLI with backup subcommand (restore stubbed)

rollroyces:
## Summary

Roadmap item: **Enterprise — backup and restore commands** (README §Roadmap). First cut: a separate `utopia` operator CLI with `backup` implemented and `restore` stubbed.

## What's in this PR

- **`crates/utopia-cli/`** — new crate, depends only on `utopia-core` for `AppConfig::load`. So `utopia backup` doesn't pull in the server, extractor, chat loop, or any of the ~50 server modules.
- **`crates/utopia-cli/src/main.rs`** — 589 lines, hand-rolled arg parser, manifest schema with sha256 checksums, `backup` shells out to `pg_dump -Fc` and `tar`, `restore` parses flags and bails with a design-doc pointer.
- **`Cargo.toml`** — workspace `members += ["crates/utopia-cli"]`
- **`.roadmap-proposals/backup-restore.md`** — design with 5 open questions

## Why a separate crate (not a subcommand of `utopia-server`)

1. The CLI doesn't need the HTTP server, the extractor, the chat loop, the mapping engine, or any of the 50+ modules pulled in by `utopia-server`. `utopia-server` is the largest crate in the workspace; making the operator compile it just to dump the database is a real CI / image-size cost.
2. The CLI is its own surface: containers, Kubernetes Jobs, cron, operators' laptops. A separate boundary makes the attack surface and the image artifact explicit.
3. The CLI reuses `AppConfig::load()` so connection info (`UTOPIA_DATABASE_URL`, `UTOPIA_DATA_DIR`, `.env`) is shared with the server. No duplicate config wiring.

## `utopia backup`

```
$ utopia backup [flags]
  --output <PATH>            archive path; default utopia-<timestamp>.tar.gz
  --include-data-dir         also tar the data/ directory
  --dry-run                  print the plan, don't write anything
  --pg-dump <PATH>           path to pg_dump binary
  --tar <PATH>               path to tar binary
  --migration-url <URL>      override connection string for pg_dump
```

What it does:

1. Resolve connection string and `data_dir` from `AppConfig`.
2. Run `pg_dump -Fc` to a temporary file under the current directory. `-Fc` (custom format) is the only format `pg_restore` consumes.
3. Optionally tar `data/` into a second temporary file.
4. Build `manifest.json` with `schema_version`, `utopia_version`, `created_at`, component paths/bytes, sha256 checksums.
[...]

PR #781, opened 2026-09-19
Add utopia CLI with backup and restore subcommands (#712)

rollroyces:
Implements the restore side of #712 (backup was the first cut) and addresses the maintainer's review notes on the open PR.

## What it does

`utopia backup` writes a single `*.tar.gz` containing a `manifest.json` (schema version, utopia build, UTC timestamp, component list, sha256 checksums) plus the `pg_dump -Fc` output and, optionally, the on-disk data directory. Refuses to overwrite an existing archive.

`utopia restore --from <archive> [--target-data-dir] [--force] [--yes]` extracts the tarball, verifies every checksum against the bytes on disk, refuses forward-incompatible manifests (schema_version greater than this binary's CURRENT), warns on older ones, then runs `pg_restore --clean --if-exists` against the target database and copies the data dir back. Without `--force`, refuses to overwrite a non-empty target data dir.

`schema_version` is the count of applied migrations in `migrations/`. A test cross-checks the constant against the directory at test time, so a future migration file without a bump to the constant fails the build instead of silently skewing forward-compatibility checks.

## Workspace

`crates/utopia-cli/` is a new crate and is added to the workspace `[members]` so `cargo check -p utopia-cli`, `cargo test -p utopia-cli`, and `cargo run --bin utopia` all work. Without this entry, the crate compiles in isolation but the workspace ignores it — the previous PR (`feat/backup-restore`) was missing this and didn't compile from the workspace root.

## Smoke test (run against the docker-compose stack)

```
$ docker compose up -d db   # pgvector/pgvector:pg16
$ sqlx migrate run
$ docker exec utopia-workspace-db-1 psql -U utopia -d utopia -c     "INSERT INTO organizations (id, name, created_at) VALUES (gen_random_uuid(), 'Smoke Org', now()) RETURNING id;"
$ # insert a matching users row referencing the org id
$ utopia backup --output /tmp/utopia-smoke.tar.gz
$ docker exec utopia-workspace-db-1 psql -U utopia -d postgres -c "DROP DATABASE utopia;"
$ docker exec utopia-workspace-db-1 psql -U utopia -d postgres -c "CREATE DATABASE utopia;"
$ utopia restore --from /tmp/utopia-smoke.tar.gz --yes
$ docker exec utopia-workspace-db-1 psql -U utopia -d utopia -c     "SELECT u.email, o.name FROM users u JOIN organizations o ON u.org_id = o.id;"
 smoke@test.local | Smoke Org
(1 row)
```

## Negative paths verified

- Forward-incompatible manifest: refuse with explicit error ("newer than this binary's 69; refusing to restore").
- Checksum mismatch: refuse with both expected and actual hashes named.
- Archive without `manifest.json`: refuse with "not a utopia backup?".
- Older manifest: warn and continue.

## Tests

`utopia-cli` 12 unit tests, all pass (was 8):

- Existing: parses, host redactor, hex encoder.
- New: manifest round-trip, schema-version policy predicate, archive-without-manifest, checksum mismatch, schema-version count cross-check.

`cargo clippy -p utopia-cli --all-targets -- -D warnings` clean.
`cargo fmt --check` clean.
`cargo check --workspace` clean.

Refs #712


WaylandYang (commented):
Thank you for coming back to this with the restore half, and for adding the crate to the workspace members so it actually compiles from the root — that was the thing that made the last cut untestable.

The shape is right: a self-describing archive rather than a bare dump, a manifest that records the schema version so that six months later a reader can tell what the file is, checksums verified against the bytes on disk before anything is applied, a refusal on a forward-incompatible manifest, and a printed plan with a `[y/N]` before `pg_restore --clean` touches anything. Backup and restore are on the schedule in [0016](docs/decisions/0016-close-the-open-seams-before-cutting-new-ones.md), so this is wanted work.

## One thing I want changed before it lands

**`--include-data-dir` puts the sealing key in the same tarball as the ciphertext it opens.**

`data/secret.key` is generated on first start and is what `utopia_core::secrets` seals credentials with — model API keys, source credentials. The database dump in the archive carries those sealed values. Anyone holding one `utopia-*.tar.gz` holds both halves, and unlike a data directory on a server, an archive is the one artifact that is deliberately copied between hosts, attached to a ticket and handed to whoever is doing the restore.

The proposal's line about encryption at rest being the filesystem's job is a fair call for the dump. This is a different thing: it is not that the archive is unencrypted, it is that the archive defeats the sealing.

What I would like:

- `secret.key` excluded from the data-dir tar by default, with `--include-secret-key` to opt in, so that including it is a sentence someone typed.
- The manifest saying whether the key is inside, so a restore can tell the operator what they are about to install, and so an archive can be audited without opening it.
- `restore` saying plainly when the key is absent: sealed credentials will not open, set `UTOPIA_SECRET_KEY` or re-enter them. That is the correct operational story for a restore onto a new host anyway — the key should come from the operator's own secret store, not ride along.

## Smaller

- **The proposal contradicts the code.** `.roadmap-proposals/backup-restore.md` still says restore "is **stubbed in this cut** — `restore` parses its flags, prints the plan, then `bail!`s". This pull request implements it. Carried over from the first cut, and it should either be updated or dropped now that the thing exists.
- **`.roadmap-proposals/` is a new top-level directory** and this repository keeps that kind of thinking in `docs/decisions`. I would rather this became a short decision record, or lived in the pull request body, than that we grew a second place where design lives.
- **`pg_dump` / `pg_restore` version skew is unhandled.** A client older than the server refuses outright, which is a confusing failure at exactly the wrong moment. The manifest records the utopia build and the schema version but not the server's Postgres version; recording it, and comparing on restore, turns a cryptic tool error into a sentence.
- **The archive is two snapshots, not one.** `pg_dump -Fc` is consistent with itself, and the data directory is tarred at a different moment, so a file uploaded between the two is in one and not the other. Not a blocker — worth a line in the manifest or the docs so nobody is surprised by a document row whose bytes are missing.

CI is green across backend, web and migrations.

WaylandYang:
The code is in good shape and I pushed one commit to your branch rather than asking you for another round. Two things I checked and one I am leaving to the maintainer.

**The design note had fallen a cut behind the code.** It said in two places that `restore` is stubbed — "parses its flags, prints the plan, then `bail!`s" and "No restore implementation (stubbed, see above)" — while `run_restore` in this same PR verifies checksums, runs the preflight, asks for confirmation unless `--yes`, and calls `pg_restore`. A note that says the opposite of the code is worse than no note, because the next reader believes it. Fixed both passages.

**What I verified here.** `cargo test -p utopia-cli` 12 pass, `cargo clippy -p utopia-cli --all-targets` clean, and `utopia backup --dry-run` against a live database prints the plan and then fails with `pg_dump not found or not executable: pg_dump. Install postgresql-client or pass --pg-dump <path>` — an explicit, actionable failure rather than a panic or a half-written archive. The connection string is redacted in that output (`postgres://***@127.0.0.1:1517/…`), which is the part I most wanted to see. I did not run `restore`, since it is destructive by design and this machine has no `pg_dump` to make an archive with.

**One thing for the maintainer, not for you.** `.roadmap-proposals/` is a new top-level directory, and CONTRIBUTING already names a home for this kind of writing: "`docs/decisions/` is where this project's reasoning lives. An ADR records why this and not that, including the approaches that were tried and failed." Your note is exactly that — it carries Options A, B and C for the image question with a recommendation. I did not move or renumber it, because which ADR number it takes, and whether the decision is settled enough to be one, is the maintainer's call rather than mine.

PR #782, opened 2026-09-19
Design: 100k benchmark first cut against real tables (#713 rework)

rollroyces:
Design document for the rework of #713. The previous cut measured Postgres filtering of a synthetic 100k-row table, which does not tell us anything about Utopia. This PR is **design only** — no code yet. Once the maintainer signs off on the shape, the code cut follows in a separate PR.

## Decisions captured

- **'document'** = `documents` row + its `chunks` + its `facts` (per maintainer's answer to Q3).
- **Write paths:** `documents::upsert_source_document_tx`, `replace_chunks`, `set_ready`, `graph::insert_fact`. Not raw SQL.
- **One scenario at a time** (per Q4). First scenario: `graph::facts_hold_at` with `at` and `as_of`.
- **Corpus size:** 100 000 documents, ~5 chunks/doc, ~5 facts/chunk = ~2.5M rows.
- **Run gating:** behind `UTOPIA_BENCH=1` (per Q5: informational, not a merge gate). No CI integration.
- **Output:** `docs/benchmarks/<utc-date>-100k.md` (per Q2).

## Where this PR leaves things

`docs/design/100k-cold-select.md` is the only file. Once maintainer signs off, the code cut adds:

- `crates/utopia-store/tests/bench_100k.rs` — fixture generator + first scenario.
- `docs/benchmarks/<utc-date>-100k.md` — the report from a real run (not in this PR; written by the bench itself).

## What I'm not asking

This is a small doc PR. If the maintainer is busy with #781 review, I'd rather this sit in the queue than force a conversation. Either way I won't start the code cut until there's a comment or 👍 on the design.

Refs #713

PR #783, opened 2026-09-19
Add 100k benchmark first cut against real tables (#713 rework)

rollroyces:
Rework of #713 per maintainer's review: the previous cut measured Postgres filtering of a synthetic 100k-row table, which does not tell us anything about Utopia. This cut populates the real `documents`, `chunks` and `facts` tables for one base through the production write functions and times `graph::entity_detail`.

## Fixture

- 1 organisation, 1 workspace, 1 knowledge base, 1 user, 1 source (`sources::create`), 1 hub entity, 1 spare entity.
- `UTOPIA_BENCH_DOCS` (default 1000) documents, each with 3 chunks via the real `chunk_text` from utopia-ingest.
- `UTOPIA_BENCH_HUB_FACTS` (default 1000) facts via `graph::insert_fact`, accumulating on the hub entity.

## Scenario

`graph::entity_detail(kb_id, hub_id, at=now, as_of=now)` — the read path a base opens when an entity is clicked. 30 iterations with 1 warmup discarded.

## Gating

`UTOPIA_BENCH=1` enables the test; without it the test prints one line and returns `Ok(())` so CI is unaffected. `UTOPIA_TEST_REQUIRE_DB` is *not* set; the test skips silently rather than failing on a database-less machine. Per maintainer's Q5 ("informational only, like 0035").

## Output

A markdown report at `docs/benchmarks/<utc-date>-100k.md`. Path resolved via `CARGO_MANIFEST_DIR` so it lands at the workspace root regardless of cargo's cwd. Includes populate-elapsed and p50/p95/p99/mean/throughput.

## Local result

```
[bench-100k] config: docs=1000 chunks_per_doc=3 facts_per_chunk=2 hub_facts=1000 iterations=30
[bench-100k] populated corpus in 6.44s
[bench-100k] wrote report to .../docs/benchmarks/2026-09-19-100k.md

| metric | value |
|---|---|
| p50 | 100.69ms |
| p95 | 105.59ms |
| p99 | 123.65ms |
| mean | 101.93ms |
| throughput | 9.8 req/s |
```

## What this PR does NOT do

- Default corpus is 1000 documents, not 100 000. Populate for 1000 docs runs in ~6s; 100k would be ~10 minutes. Operators pass `UTOPIA_BENCH_DOCS=100000` for the maintainer's target.
- Does not add additional scenarios. Each future scenario (hybrid retrieval, reconcile_new_fact, paging, deletion) is its own PR per the design in the closed #782.

## Design doc

The design rationale lives in #782 (closed as superseded). The cut resolves every answer to maintainer's 5 questions concretely: real tables, dated report, real read path, one scenario, env-gated.

Refs #713


WaylandYang (commented):
This is the rework I asked for, and it is a different thing from the first cut. Timing `graph::entity_detail` over rows written by `sources::create`, the real `chunk_text` and `graph::insert_fact` measures a read path this product actually has; the earlier version measured Postgres filtering a synthetic table, which would have been true of any schema.

What I like:

- **The scenario is a real gesture.** `entity_detail(kb_id, hub_id, at, as_of)` is what a base does when someone clicks an entity, and the hub with a thousand facts on it is the shape that gets slow first.
- **p50, p95 and p99 rather than a mean alone.** A mean over 30 iterations would have hidden exactly the tail this is worth running for.
- **Gating that cannot fire by accident.** `UTOPIA_BENCH` unset prints one line and returns; no database returns quietly. CI stays untouched, which is the right call for an informational bench, same as 0035.

## Two things to settle before it lands

**1. The fixture stays in the database.** A run writes about 2.5 million rows into a base called `bench-100k` and nothing removes them. That is defensible — you often want the data still there to re-measure the read without paying the load again — but right now it is undocumented and there is no reuse path either, so a second run silently adds a second base of the same name and the operator's database grows by 2.5M rows a time. Pick one and say it in the module comment:

- drop the base at the end unless `UTOPIA_BENCH_KEEP=1`, or
- reuse an existing `bench-100k` base when one is there, and say how to drop it.

**2. The number has no context to be read against.** A p95 on someone's laptop, against a cold shared buffer pool, is a number that cannot be compared with the one the next person posts. Two lines in the report would fix it: the Postgres version and `shared_buffers`, and whether the run was warm. The bench already discards a warmup iteration, so the second is nearly free to state.

## Smaller

- `UTOPIA_BENCH_DOCS` defaults to 1000, which the header calls "100k" through 3 chunks and a thousand hub facts. The name is fine as a label for the scale, but the module comment should say what the default actually loads, since "100k" and "~2.5M rows" both appear and a reader will not know which knob produces which.
- Consider printing the fact count on the hub next to the latency. When this is re-run in six months against a different default, that pairing is what makes the two numbers comparable.

CI is green. I have not run it here — that is the point of the gate — so the numbers in the description are yours, not mine.

WaylandYang:
Reworked exactly as the review asked, and the numbers come from the real read path, so this goes in. I pushed one commit to your branch rather than sending you round again.

**The bench took its corpus out with it.** `kb()` creates an organisation, a workspace, a base, a user and a source, and `populate_corpus` fills them; nothing removed them afterwards. At the default that leaves a thousand documents, three thousand chunks and several thousand facts in whatever database the runner points at, and at the `UTOPIA_BENCH_DOCS=100000` setting the doc calls for, about 2.5 million rows. A second run stacks another corpus on the first, which also moves the number being measured, and every other database-backed test in this repo shares that database.

One statement does it, because every foreign key on the path is `ON DELETE CASCADE`:

```rust
sqlx::query("DELETE FROM organizations WHERE id = $1").bind(_org_id).execute(&pool).await?;
```

It runs right after `samples.sort()`, not after the report is written: a report that fails to write should not leave the corpus behind either.

**The table rows were missing their closing pipe.** The header had one and the data rows did not, so the file read as a table with a ragged edge. Four lines.

Checked here: `UTOPIA_BENCH=1 UTOPIA_BENCH_DOCS=30 UTOPIA_BENCH_HUB_FACTS=50 UTOPIA_BENCH_ITERATIONS=5` against a local database passes, the report renders, and `SELECT count(*) FROM organizations WHERE name='bench-100k'` is 0 afterwards. `cargo fmt --all --check` clean.

Two things I am leaving as they are, for the record rather than for you to change: the test writes into `docs/benchmarks/` inside the working tree, which is unusual for a test but gated behind `UTOPIA_BENCH` and documented; and a panic in the middle of the measurement still leaves the corpus, since the delete is not in a guard. Neither is worth more machinery for a bench that only runs when someone asks for it.

PR #790, opened 2026-09-19
Render nested proof trees in the entity panel (#477 cut 3)

rollroyces:
Cuts 1 and 2 of #477 landed as record 0030 + the runner in #485; the backend's `proof()` now returns `ProofStep` with a recursive `premises: Vec<ProofStep>` field. The frontend type was still flat and the renderer treated the result as a single `<ol>`, which works for proofs whose premise tree is one level deep and silently underflows for anything deeper.

## What this changes

- `web/src/api.ts`: `ProofStep` gains `premises: ProofStep[]`, mirroring `utopia_core::models::ProofStep`. Comment now says 'tree, not chain' (0030's wording).
- `web/src/pages/Graph.tsx`: `ProofSteps` is unchanged at the API level (still takes `steps: ProofStep[]`). Internally it delegates each row to a new `ProofStepRow` component, which renders one step plus, recursively, its own `premises` as a nested `<ol>`. Recursion bottoms out at the server's depth cap; leaves have `premises.length === 0`.

Both call sites — `derivedProof` on a landed derivation and `blockedProof` on a violation that blocked one — get the recursion without further change.

## Visual treatment

The nested `<ol>` gets `ml-4 border-l border-edge pl-3` so a multi-level proof reads as a tree visually, not as one tall column. This is the existing convention for indented children elsewhere in the same file (`ExpandCard`, the recursion in `MapPane`).

## Verification

```
pnpm typecheck   # tsc --noEmit, clean
pnpm test        # 82 vitest tests, all pass
pnpm build       # vite build, clean
pnpm guard       # style-guard, 84 files compliant
```

Refs #477


WaylandYang (commented):
The diagnosis is right and the shape of the fix is right. `proof()` has returned a recursive `ProofStep` since #485, the frontend type was flat, and a renderer that maps one `<ol>` over the top level shows a two-level proof as if it were one level — which is worse than showing nothing, because the reader cannot tell that a premise had premises of its own. Making the type mirror `utopia_core::models::ProofStep` and letting the component recurse into `premises` is the smallest correct change, and stopping when `premises` is empty is the right base case: a leaf's evidence is the sentence, and there is nothing further to ask.

Comments read well too — "tree, not chain" is 0030's own wording, and saying that the depth limit belongs to the server rather than the client is the part someone would otherwise re-litigate.

## What I could not check, and what would fix that

**There is no test, and there is no way to see it.** The pull request touches `api.ts` and `Graph.tsx` and nothing else. I went looking for a nested proof to render and there is none in any base I have: every base here reports zero facts with `derived_by_rule`, so the recursive path cannot be exercised by clicking. The web suite and the production build pass on the branch, which tells me it compiles, not that it nests.

`web/src/pages/Graph.test.ts` already exists and already imports a function out of `Graph.tsx`, so the road is open: export `ProofSteps`, or lift the flattening into a small pure function beside it, and assert on a three-level fixture that each level's `<ol>` sits inside its parent's `<li>` and that a leaf renders no list at all. Ten lines, and then the thing that this pull request exists to fix has something holding it in place.

That is the one change I would like before it lands. The rest I am happy with.

## One question

A deep tree in a narrow panel: does anything cap the indentation, or does level six sit against the right edge? The server caps depth, so there is a bound, but the panel is 384 wide at its widest. If you have a fixture handy, a screenshot of the deepest proof the engine will produce would answer it faster than either of us reasoning about it.

Issue #268, opened 2026-09-03
Deleting a document leaves its facts in the graph with no evidence


WaylandYang:
**Measured** on an isolated database — one document, one chunk, two entities, one fact with evidence, then `DELETE FROM documents`:

```
before  docs=1 chunks=1 entities=2 facts=1 evidence=1
after   docs=0 chunks=0 entities=2 facts=1 evidence=0 invalidated=0
```

`documents::delete` (`documents.rs:400`) is a bare `DELETE FROM documents`; everything else is FK cascade. `chunks.document_id` and `fact_evidence.document_id` / `fact_evidence.chunk_id` are all `ON DELETE CASCADE`, so every evidence row goes. `facts` has no FK to documents and is `append-only，永不 DELETE`, so the fact stays — live, unflagged, and now unsourced.

**Three things are wrong with that.**

1. **The product says otherwise.** The Privacy policy's "Retention and deletion" section (`web/src/i18n/en.ts:314`) and the delete confirmation dialog (`en.ts:618`) both promise "Facts already extracted into the graph remain, **with their provenance**". The provenance does not remain.
2. **It contradicts the house rule everywhere else.** `replace_chunks` soft-deletes superseded chunks rather than removing them, for the stated reason that `fact_evidence` references must not break (`documents.rs:415`, `migrations/0002_ingest.sql:103`); `fact_evidence` is introduced as 溯源一等公民 (`0003_graph.sql:298`), and its `chunk_id ... ON DELETE CASCADE` is the one FK in that file with no justification comment. Rejecting a fact keeps its evidence because deleting it "would erase what we once believed" (`graph.rs:1523`). An account is deactivated, not deleted, so the ledger stays attributable. Document deletion is the only place that subtracts.
3. **The blob is never deleted.** `BlobStore` (`blob.rs:15`) has no delete method and nothing in the repo calls one, while `document_versions` rows cascade away — so the original file stays on disk, unreachable and unreplayable. Nothing today can remove it.

A fourth follows from the first: after a hard delete, the next sync of a source-backed document finds no row for its `external_key` (`documents.rs:189`), creates a fresh document and extracts again, while the old facts were never invalidated — **the same assertion lands twice**.

**Decision.** Deleting a document is an event on the recording axis. Delete leaves a tombstone and retires what depended on the document; purge is the separate, explicit step that removes content.

*Delete*
[...]

PR #791, opened 2026-09-19
Restore shared facts when either deleted source returns

Maya-Kid:
When two documents support the same fact, deleting A keeps it alive and deleting B retires it. Restoring A brings back its document and chunks, but the fact remains retired because only B's deletion record lists that fact. If B was purged, restoring B is no longer an option either. This affects ordinary document undo and the source-sync/re-upload paths that share `restore_tx`.

Include facts cited by the restored document when an unreverted deletion record identifies their retirement and their invalidation timestamp still matches that deletion transaction. Include these candidates in the existing timeline lock/recalculation, and compare their timestamps again before reviving them. Facts retired before the deletions and facts supported only by the still-deleted document stay retired. Existing handling of the restored document's own deletion record remains unchanged. No schema, dependency, or HTTP API changes.

Regression tests call the real store deletion, purge, and restore functions against PostgreSQL. On dev `87df718`, both new cases fail while the existing deletion test passes. The new cases cover restoring the first source with the last source either soft-deleted or purged; each also checks that unrelated and previously retired facts stay retired.

This addresses sequential delete/restore of shared evidence; it does not claim to serialize all concurrent document lifecycle operations. It follows the existing #268 distinction between retirement due to document deletion and an independent withdrawal.

Validation: Linux `cargo test --locked --workspace` reports **893 passed, 0 failed, 1 ignored**, with `UTOPIA_TEST_REQUIRE_DB=1` and an isolated PostgreSQL/pgvector database. The ignored test requires public HTTPS; optional real-provider tests and the PDF fallback executable are not configured. Workspace clippy (`--all-targets -- -D warnings`, Rust 1.98.1), fmt, and diff checks pass. No production data or live model endpoint was used.


WaylandYang (commented):
Verified. On the branch all three pass; with only `documents.rs` reverted and your tests kept:

```
test restoring_either_source_recovers_a_shared_fact ... FAILED
test restoring_a_source_recovers_a_fact_after_the_last_source_was_purged ... FAILED
test a_deletion_is_an_event ... ok
```

The existing deletion test staying green is the half that matters most — a revival rule is one wrong predicate away from bringing back things that were retired on purpose.

The predicate is the right shape. A fact comes back only when it is cited by a chunk of the document being restored, is currently invalidated, appears in an unreverted deletion record, **and** its `invalidated_at` equals that deletion's `deleted_at`. That last clause is what keeps a fact retired for some later reason from riding back in on an old deletion record, and re-checking the `(id, invalidated_at)` pair under the timeline lock closes the window between the read and the write. Your two negative cases — a fact retired before the deletions, and one supported only by the document that is still deleted — are exactly the two I would have asked for.

The purge case is the one I would not have thought to cover, and it is the one that makes this worth landing: once B is purged there is no second undo to reach for, so without this the fact is simply gone while its evidence is live again.

Nothing to change.

PR #775, opened 2026-09-19
Reconsider negative bindings when existing definitions change

Maya-Kid:
Editing an existing class or property definition leaves automatic `none` / `undecided` bindings out of the alignment work list. Both negative stale predicates now compare against the newest `updated_at` in the same KB, so an edited definition can be considered again. Bound results still depend only on their selected definition, and the existing person-decision protections are unchanged.

Fixes #773.

The database regression covers both negative statuses for classes and properties, new candidates, bound-definition edits, unchanged inputs, KB isolation, fresh negative decisions returning to cache, and person decisions rejecting automatic writes. Fixture timestamps establish ordering without sleeps; signatures and edits use the real store APIs. This tests the store lifecycle, not a complete model-backed worker run.

Validation: Linux x86_64, Rust 1.98.1, PostgreSQL 16.15 / pgvector 0.8.6, with `UTOPIA_TEST_REQUIRE_DB=1` and an isolated database. Final regression: 6 passed (18 lifecycle cases); the old predicates fail the four automatic-edit regressions and the newly-stale person-decision control. Full workspace: 875 passed, 1 existing live-HTTPS test ignored, 0 failed. Also passed: `cargo fmt --all --check`, `cargo clippy --locked --workspace --all-targets -- -D warnings`, `cargo test --locked --workspace` with DB and PDF requirements enabled, `cargo build --locked --workspace`, and the frozen-lockfile frontend install, 63 frontend tests and production build.

This is deliberately the small timestamp-based fix. Cosmetic edits also update these timestamps and can cause extra reevaluation. Changes during an in-flight model call still need input-version tracking; this PR does not change that concurrency contract or introduce migrations. The review feedback on reevaluation cost is addressed in the design page: debouncing reduces runs, not the number of items reconsidered; two votes per eligible item are made through batched requests.

Tested commit `c33e20c3a4ebb62ccbcee85ca2f746af6190fe64` on top of `dev@966b4028f6630117ec9b623f17d6e9576880ac68` (includes #772). Linux source hashes match every tracked file in this commit.

Documentation-only follow-up `c0983b2` addresses the review request; `git diff --check` passed. The implementation and regression tests are unchanged from the Linux-tested commit above.

PR #792, opened 2026-09-19
Keep kind-word decisions and entity types consistent

Maya-Kid:
Kind-word decisions and entity types can disagree in two cases: re-evaluation changes a bound signature to `undecided` but leaves its old aligned class on entities; or a person chooses `none` after an agent saves its binding but before the agent applies its class. In the latter case the final binding belongs to the person, while the entity receives the agent’s older class.

Store each accepted decision and its entity projection in one transaction, shared by the type-alignment worker and the human review handler. The binding row remains locked until the entity update completes. Both `none` and `undecided` retract aligned types, while explicit human entity classifications and the existing person-decision precedence remain protected. Graph listeners are notified after accepted changes. Model calls run before this short write transaction.

Four regression tests exercise:
- Disagreement retracting an old aligned type, including the downstream phrase signature and an explicitly human-classified control entity.
- A human binding decision committed while the model response is held at a notification barrier.
- A human `none` decision racing the agent’s entity write, ordered by a real PostgreSQL row lock and observed blocker chains, without timing sleeps.
- Agreeing bound/none votes retaining their behavior.

Validation on Linux, Rust 1.98.1 / PostgreSQL 16.15, with required database tests and an isolated build directory:
- Original dev fails the disagreement regression; the other two original controls pass.
- The non-transactional decide/apply sequence fails the new row-lock regression; the transaction fixes it. All four targeted tests pass in the final workspace run.
- Formatting, workspace Clippy (`-D warnings`), and workspace tests pass: **895 passed**, one existing live public-HTTPS test ignored.
- Frozen frontend install, 82 frontend tests, and production build also passed.

This follows #775. It applies when a binding is evaluated/decided and does not backfill previously cached undecided projections. The separate in-flight definition-version issue is tracked in #795.


WaylandYang (commented):
Verified, and the rule that matters most survives: a person's decision still cannot be overwritten by the agent, now by two mechanisms rather than one. The `WHERE NOT (decided_by = 'person' AND EXCLUDED.decided_by = 'agent')` clause still rejects the write, and `decide_and_apply` only runs `apply`/`unapply` when the write actually happened, so a rejected agent decision now provably touches neither the binding nor the entities. That second half is the part the old code was missing, and it is the bug.

Reverting the three source files and keeping your tests:

```
assertion `left == right` failed: an undecided binding must not leave an aligned class on its entities
  left: Some(01a0b945-5170-70e1-bafd-40fa2cbd9d7a)
 right: None

assertion `left == right` failed: the older agent must not undo the person's none decision
  left: Some(01a0b945-535c-7642-bfdd-d2a07de05175)
 right: None

test result: FAILED. 2 passed; 2 failed
```

Six runs each way, no flake in either direction, which is unusual for a test that races two transactions and is worth the credit.

Two of the four are guard tests rather than regression tests — `agreed_votes_keep_their_existing_behavior` and `human_decision_during_disagreement_survives` pass before and after. That is correct for their purpose; the body should not be read as four tests pinning the bug, because two do.

## One thing to add, and one to think about

**Your tests self-migrate and the others do not.** The fixture calls `db::migrate`, so on a database that already has the schema all four fail with `relation "organizations" already exists`, while #793 and #794 run fine on the same database. I had to make a second database to run them. CI is fine either way; the inconsistency is what will cost somebody an afternoon.

**The lock is now held across an unbounded entity update.** The transaction holds the `type_bindings` row from `decide` to commit, and inside that it updates every entity in the base whose normalized `specific_type` matches the kind word. For a word like "company" that is most of the base. Two consequences:

- `ON CONFLICT DO UPDATE` locks the conflicting row before it evaluates its `WHERE`, so even a *rejected* agent decision briefly blocks a person on the same word. Harmless as written, because you skip `apply` when nothing was written and that transaction is short.
- `decide_alignment_kind_word` is a synchronous HTTP handler and the main pool sets no `lock_timeout`. A person's click can now wait behind the agent's full entity update with no bound. That is the correctness trade you chose and I agree with it, but the handler should set a `lock_timeout` and answer 409 rather than hang. I do not need it in this pull request; I would like an issue.

No new deadlock. `decide` is the only production writer of `type_bindings` in the workspace, so the order is always bindings then entities and nothing takes them the other way; `align_types` already serializes agent runs per base with an advisory lock, so the only contender for a row is a person through the review route.

Smaller: the `propose` call in the `none` branch stays outside the transaction, so "one transaction" does not cover the proposal; and `decide`, `apply` and `unapply` stay `pub`, so a later caller can still write one without the other. A line of doc comment on each, or `pub(crate)`, would keep that from happening by accident.

PR #793, opened 2026-09-19
Keep Unicode source filenames from aborting synchronization

Maya-Kid:
A long mixed ASCII/Chinese title can panic during filename construction: the 60-byte `String::truncate` cutoff may fall inside a UTF-8 character. RSS feed-only imports, GitHub issues and Jira issues use this helper. The panic aborts the source run before later entries are imported, and retrying encounters the same title again. URL filenames have the same problem at their 120-byte cutoff.

Use the existing `truncate_utf8` helper in both places, preserving the byte limits and ASCII behavior. Full titles and document content are unchanged.

Validation on dev `87df718`:

- The real `sync_source` path consumes a loopback RSS feed with a mixed-language title followed by another article, using PostgreSQL. Before: `is_char_boundary` panic. After: both documents are persisted and the source completes with status `ok`.
- A direct regression covers title and URL filename boundaries plus unchanged ASCII truncation.
- Linux workspace tests report **893 passed, 0 failed, 1 ignored**, with `UTOPIA_TEST_REQUIRE_DB=1`. Workspace clippy, fmt and diff checks pass (Rust 1.98.1). The ignored case needs public HTTPS; optional external services and the PDF fallback tool are not configured.

No live GitHub/Jira endpoint or production data was used. No dependencies, schema, or public API changes.


WaylandYang (commented):
Verified, and the bug is worse than your title says.

Reverting the source and keeping your tests:

```
panicked at crates\utopia-server\src\ingest_sources.rs:902:7:
assertion failed: self.is_char_boundary(new_len)
test result: FAILED. 0 passed; 2 failed
```

Both deaths land at `slugify`, so I reverted only the 120-byte half on its own to make sure that one is independently covered:

```
panicked at crates\utopia-server\src\ingest_sources.rs:448:10:
assertion failed: self.is_char_boundary(new_len)
test result: FAILED. 1 passed; 1 failed
```

Both halves are pinned. The two you fixed were also the only `String::truncate` calls in `crates/` — the other six are `Vec::truncate` — so the sweep is complete.

**The part worth adding to the body: this does not just abort a run, it wedges the source.** A panic is not an `Err`, so `sync_source`'s `match outcome` never runs and neither does `finish_sync`, which leaves `sources.last_sync_status` at `'running'`. The job worker catches the panic through the `JoinError`, so the job is marked failed and the operator sees something — but `due_sources` filters `WHERE last_sync_status NOT IN ('queued','running')` and `mark_queued` has the same guard, so the scheduler never picks that source up again and the manual sync route quietly does nothing. One Chinese title with an unlucky byte offset and the source is off until someone edits the row by hand. Each retry also leaves another orphan `'running'` row in `source_sync_runs`.

That is the sentence that makes this a fix worth reading about six months from now.

Two notes:

- The blast radius is wider than RSS: `slugify` also feeds the GitHub and Jira readers. Your unit test covers the helper directly, so all three are covered; the integration test exercises RSS only, which is fine.
- Adjacent and not yours: `ensure_extension` caps nothing, so custom, GitHub and Jira titles produce unbounded filenames. Not a panic, not a regression.

Full server suite on your branch: 283 passed against the base's 281, with the same 17 environment failures on both sides.

PR #794, opened 2026-09-19
Keep failed sync attempts from skipping incremental source updates

Maya-Kid:
A failed source sync advances `sources.last_sync_at`, and the custom, GitHub and Jira readers reuse that attempt timestamp as their next incremental lower bound. If the endpoint returns 503, its next successful response can exclude every update preceding that failure: the source reports success with no imported documents. Using the end time also skips changes made while a successful run was fetching its snapshot.

Derive the incremental lower bound from the start of the most recent successful `source_sync_runs` record, and pass it explicitly to those three readers. Keep `last_sync_at` unchanged for scheduling and attempt diagnostics. If no successful record remains in the bounded run history, omit the lower bound and replay; existing document identity/content deduplication applies. No schema, dependency, or HTTP API changes.

Three regressions exercise `sync_source`, loopback HTTP and PostgreSQL. The fixture filters its response by the actual `since` query parameter:

- a failed first sync must still fetch existing items on retry;
- a failure after an earlier successful run must not move its checkpoint;
- an update within the preceding successful run's time window must remain eligible even without an intervening failure.

All three fail on dev `87df718` with zero imported documents instead of one. The failed-attempt timestamp/status remain recorded, preserving scheduler behavior.

This prevents future timestamp-related gaps. It does not automatically repair updates already skipped before this fix, change pagination/per-run caps, or claim live-provider acceptance. Previously missed history may require a deliberate full replay.

Validation: Linux `cargo test --locked --workspace` rerun reports **894 passed, 0 failed, 1 ignored**, using isolated PostgreSQL/pgvector with `UTOPIA_TEST_REQUIRE_DB=1`. Workspace clippy, fmt and diff checks pass (Rust 1.98.1). Optional real-provider tests and the PDF fallback tool are not configured; the ignored test requires public HTTPS. No production data was used.

The first full run hit a deadlock in the existing `evidence_follows_a_row_rewritten_while_it_was_written` test. PostgreSQL logs identify the other participant as the global `UPDATE facts` from the concurrent migration-backfill test. The unchanged full suite passed on rerun; this PR does not alter that test or the temporal code.


WaylandYang (commented):
The bug is real and the fix is the right one: `finish_sync` writes `last_sync_at` on the error path too, so the next incremental read starts after a failure that fetched nothing. Reverting the source and keeping your tests:

```
Error: incremental cursor skipped an unimported update: 0 documents
test result: FAILED. 0 passed; 3 failed
```

All three, for the right reason. And `updates_during_a_successful_sync_remain_in_the_next_window` fails on the old code with no failure involved at all — it catches a **second** bug, that a change landing while a successful run was fetching its snapshot was skipped because the cursor was the run's finish time. That is a widening past the title and worth saying out loud in the body.

## One gap I filled on your branch

Your tests assert that the missed update arrives, and the fixture returns it when `since` is `None`, so I mutated the branch to pass `since = None` always — incremental reading disabled entirely — and:

```
test result: ok. 3 passed; 0 failed
```

All three still passed. A regression that turned every custom, GitHub and Jira sync into a full fetch would have gone through untouched.

I pushed a commit that records the bound the fake server actually received and asserts it: with a prior success it must be within a second of that run's `started_at` and strictly before the change, and with no prior success it must be absent. With the same mutation applied, two of the three now fail with `没有带下界：增量拉取整个没了`, and the third correctly stays green because a source that never succeeded should not send a bound.

## Two things for the body rather than the code

- **Fifty consecutive failures evict the last success.** `finish_run` trims to the 50 most recent runs per source, so `max(started_at) WHERE status='ok'` can go empty and the next attempt replays from nothing. Reproduced against the real SQL in a rolled-back transaction: `before trim: cursor 2026-06-11 10:41:47+00` then `after trim: cursor NULL -> full replay`.
- **That replay does not fully recover a large source.** All three readers cap at `MAX_NEW_PER_SYNC = 200` with a silent `.take(200)` and still finish `'ok'`, so a source with more than 200 changed items gets 200 back and the cursor advances past the rest. Still strictly better than the old behaviour, which skipped everything — but "omit the bound and replay" reads like a clean recovery and it is not.

The deduplication you lean on is there: `ingest_item` resolves `find_by_external_key` and compares `sha256`, so an unchanged item returns `Unchanged` with no new version, no blob write and no extraction job. A replay costs network, not the graph.

Full server suite on your branch: 284 passed against the base's 281, same 17 environment failures on both sides.

PR #797, opened 2026-09-19
Prevent overlapping materializations from duplicating typed facts

Maya-Kid:
Two overlapping typed-graph recomputations can create two live typed facts from one open statement. Both the alignment worker and a human phrase decision call `materialize`; each run can read the same due statement and see no existing typed fact before either inserts it.

Serialize recomputation per knowledge base with a transaction-scoped advisory lock acquired before reading the work. Perform the complete recomputation on that transaction’s connection. The graph insertion path and temporal rewrites accept the same connection, so a waiting materializer does not hold a pool slot while the active one needs to acquire another. Temporal subtransactions use savepoints; the existing insertion rules are reused.

The regression uses a fixture-scoped PostgreSQL trigger as an INSERT barrier and observes `pg_blocking_pids` to order the two real materializer calls. Original dev produces **2 live facts** and fails; the fix produces **1** and passes. It also covers a two-connection pool, a single-connection rerun, a temporal closing rewrite on that single connection, and idempotent reruns. There are no timing sleeps or model calls.

Validation on Linux (Rust 1.98.1, PostgreSQL 16.15, isolated database/build directory): formatting, workspace Clippy with `-D warnings`, and the full workspace suite pass (**897 tests**, one existing live public-HTTPS test ignored). Database and PDF fallback tests were required. Frozen frontend install, 82 frontend tests, and production build also pass.

Recomputations for one base now wait for each other; different bases have separate lock keys. This prevents new concurrent duplicates and does not rewrite previously duplicated historical facts.


WaylandYang (commented):
The bug is real and the test discriminates it. Reverting all three source files — the test compiles unchanged, no shim needed — gives, three times out of three:

```
Error: overlapping runs produced 2 live typed facts; expected one
test result: FAILED. 0 passed; 1 failed
```

Building the barrier out of a fixture-scoped trigger and `pg_blocking_pids` rather than a sleep is the right way to write this, and it is why it runs in 0.4 seconds.

**One consequence is better than the description claims.** Running the whole recomputation on one transaction makes it all-or-nothing. Forcing a failure on the second of two statements:

| | typed facts | sources |
| --- | --- | --- |
| before | 2 | 1 |
| this branch | 0 | 0 |

The old code left a typed fact with no source row for the next run to clean up. Worth saying.

The savepoints are fine, and I checked what they actually do: `ROLLBACK TO SAVEPOINT` then `ROLLBACK`, because nothing catches the error, so the outer run aborts rather than continuing. They exist to make `begin()` legal on a connection already in a transaction, not to isolate a failure. No caller relied on the inner transaction committing independently — the other callers of the three `temporal::` functions still pass `&pool` and still get a real transaction.

No lock-order inversion: `align_types` and `align_phrases` both use `try_` variants, so neither can be the waiting edge of a cycle, and materialize never takes their keys.

## What I want changed, and it is not the lock itself

**A materializer waiting for the lock pins a pool connection, and the wait has no bound.** Two waiters on a two-connection pool, then a third checkout:

```
PROBE blocked backends = [("Lock","advisory","SELECT pg_advisory_xact_lock(hashtext('typed_materialize'), …")
                          ("Lock","advisory","SELECT pg_advisory_xact_lock(hashtext('typed_materialize'), …"]
PROBE third acquire after 3.0063752s => Some("pool timed out while waiting for an open connection")
```

`acquire_timeout` covers checking out a connection, not waiting on the lock, so the wait itself is unbounded. And the run is not short: 500 statements takes

```
first run 4.2545363s -> Outcome { retired: 0, added: 501, merged: 0 }
```

so the first bind on a large base is minutes of exclusive per-base lock.

Now the part that matters: `decide_alignment_phrase` (`review_routes.rs:1280`) calls `materialize` **inline, inside the HTTP request, with no lock of its own**. The worker path is serialized by `align_phrases`'s try-lock, but review decisions are not, so any number of people deciding phrases in one base queue up on this lock, each holding one of 32 connections, each request hanging until the run in front finishes.

The lock is right. The inline call from a request handler is what should go. Two ways, and you know this code better than I do now:

- the handler enqueues instead of materializing, and the worker's existing end-of-run recheck picks up what arrived late — that pattern is already in `phrase_alignment`, added in #757 for exactly this reason; or
- the handler takes `pg_try_advisory_xact_lock` and returns "already recomputing" when it cannot, which is honest and instant, again leaning on the recheck.

I am merging this as it stands, because duplicate typed facts are worse than a slow decision, and filing the handler question separately.

## Smaller

- **The row-lock footprint is wider than the title.** Every typed fact the run rewrites is now locked until commit: `FOR UPDATE NOWAIT` from another session returns `could not obtain lock on row in relation "facts"` on this branch and succeeds on `dev`. `graph::add_evidence_located` takes `FOR SHARE`, so ingestion now waits for a whole materialization run rather than for one rewrite. No deadlock reproduced, and I would not expect one, but it is a change worth naming in the body.
- **The test leaks on panic.** It creates a function and a trigger on the shared `facts` table and cleans up on `Err` but not on a panic — `signatures(...).remove(0)` on an empty vec, or a `JoinError` from `a.await??` — which would leave a trigger in a shared database. A guard that drops them on unwind would cost little.
- `materialize.rs:8` still says writes go through `insert_fact` / `insert_value_fact`. This pull request edited that line and removed both calls.
- The duplicate class is not closed in general: `insert_fact_inner`'s read-then-insert is still unguarded for `pending.rs` and `names.rs`, and there is no unique index on `facts` enforcing one live typed fact per statement. Honest to the title, worth knowing.

`utopia-store` fully green, `utopia-server` 300 passed, clippy and fmt clean.

Issue #795, opened 2026-09-19
An ontology edit during type alignment can be hidden by the response timestamp


Maya-Kid:
This tracks the separate in-flight input-version gap noted in the review of #775. I reproduced it against `dev` (`87df718`) on Linux with PostgreSQL 16.15 and a scripted HTTP model; no real model or timing sleeps are involved.

A definition edited while type alignment is waiting for the model can be treated as already considered, even though neither vote saw the new definition.

Reproduction:
1. Start with one class whose description is `OLD definition` and one entity with kind word `company`.
2. Run the real `align_types` worker. After the mock endpoint receives the first request, it commits a class edit (`description = 'NEW definition', updated_at = clock_timestamp()`) **before returning the response**.
3. Return `none` for both votes. Both captured requests contain `OLD definition`, not the updated description.
4. The binding's `decided_at` is later than the class edit, `type_bindings::stale` returns nothing, and a second worker run sends no new model requests.

The desired boundary is that a result based on an older definition should either be discarded/requeued or remain detectably stale. Merely widening the negative-cache predicate, as #775 does, cannot catch an edit hidden behind the later response timestamp.

A standalone [reproduction branch](https://github.com/Maya-Kid/utopia/tree/codex/alignment-input-version-repro) adds only a test module to the dev baseline. With an isolated migrated pgvector/PostgreSQL database (the fixture also runs migrations):

```sh
UTOPIA_DATABASE_URL=postgres://... \
UTOPIA_TEST_REQUIRE_DB=1 \
cargo test -p utopia-server version_repro -- --nocapture
```

It passed in the isolated Linux build. **Passing means the assertions reproduced the current incorrect freshness result**, not that the gap is fixed; this test-only branch is not a proposed mergeable fix.

I have limited the dynamic claim to type alignment. This is a design follow-up for choosing the input-revision contract, per CONTRIBUTING, rather than a request to expand #792 or change the schema without agreement. Human decisions should retain their existing precedence.

Maya-Kid:
Moving the design discussion from #804 here as requested in review, and closing that PR. This is a proposal for agreement, not an accepted contract or implementation.

The proposed invariant is that an automatic decision can be current only for the inputs it actually read. A reply timestamp cannot establish that. A possible protocol is:

1. Read the input revision and all consumed definitions/hierarchy in one short, consistent snapshot transaction; close it before model calls.
[...]

PR #801, opened 2026-09-19
Respect inherited phrase candidates and retire unsupported bindings

Maya-Kid:
Phrase alignment excludes subclasses from properties declared on their ancestors. When a property edit removes every candidate from an existing automatic binding, it also leaves the old binding and typed projection in place.

Follow all parent edges within the knowledge base when filtering candidates, and show those ancestors and the direction of inheritance to the model. Record structural absence as `none` and use normal materialization to retire unsupported projections, preserving human decisions, other supporting sources, original statements and quotes. Candidate overflow remains distinct from absence.

Stale work is restricted to signatures still backed by live open statements. This takes the live-signature alternative suggested in review: when kind-word classification changes a signature’s endpoints, its orphaned binding no longer queues work forever. Binding rows themselves are retained; no human decisions are deleted. No migration, dependency or additional query round trip is introduced.

The review revision also guards CI against running zero database worker tests, removes the redundant assertion and structurally impossible cross-KB fixture, cleans up the injected trigger after worker errors/panics/timeouts, and clarifies the closure and cache documentation. It removes the assertion that treated overflow requeue as a desired outcome, without changing overflow scheduling.

Validation on Linux (Rust 1.98.1, PostgreSQL 16.15/pgvector 0.8.6): formatting, workspace Clippy with warnings denied, build, and 932 tests passed; one existing live-network test is ignored. All 14 focused tests passed, including 12 real-database worker tests using a scripted local model. Against the preceding PR implementation (9edb707), the revised tests fail on orphan requeue and missing hierarchy context; the revised implementation passes both. The CI guard accepts the worker log and rejects zero-test and unit-only logs. No live-provider quality evaluation is claimed.

Remaining limitations:
- Parent-edge edits alone do not invalidate cached decisions: adding a parent does not reopen a negative, and removing one does not retire an inherited binding/projection. The regression uses a property edit to trigger reevaluation. Hierarchy and in-flight input freshness remain separate (#795).
- Stale over-limit bindings may still repeatedly queue, and the existing final scheduling window is unchanged. This revision does not claim that scheduling problem is resolved.
- Unresolvable model property keys remain unchanged. Worst-case ancestry closure storage can be quadratic.

Run the focused suite against a dedicated test database:
```sh
UTOPIA_DATABASE_URL=postgres://... UTOPIA_TEST_REQUIRE_DB=1 cargo test -p utopia-server phrase_alignment:: -- --color never
```


WaylandYang (commented):
Both bugs are real and seven of your twelve worker tests catch them. Reverting `phrase_alignment.rs` and keeping the tests:

```
test result: FAILED. 6 passed; 7 failed
left: "bound" right: "none"          (stale automatic bound must be retired)
left: "none"  right: "bound"         (a property edit reopens a negative subclass binding)
subclass manufacturer missing from actual worker request
```

Computing the ancestor map once per run out of the classes already fetched for `class_key`, at zero extra queries, is the right place for it. I measured the closure on shapes that exist: 900 classes at depth 6 is 0.16 MiB and 0.6 ms, two parents each is 0.26 MiB and 1 ms. The quadratic in your scope note only bites at depth 2000 (54 MiB) and 5000 (347 MiB), which nobody hand-authors. That is not what worries me.

## What blocks it: this creates an `align_phrases` loop that never converges

Writing `none` for a structurally impossible signature leaves a row behind, and rows outlive signatures. `phrase_bindings::stale` selects from `phrase_bindings` with **no join to the live signature set**, so once a signature key changes the old row is stale forever: the worker cannot re-decide it, because there is no signature to decide, so `decided_at` never advances, and the tail check `stale(...).any(|b| b.decided_by != "person")` re-enqueues the job on every run.

The trigger is the most ordinary sequence in this pipeline — a phrase seen before its subject's kind word is bound, which the module header calls the normal early state, and then the kind word lands and the key changes:

```
# this branch
AUDIT round 1: stale=1 queued_jobs=1 rows=[(None,"none"), (Some(...),"bound")]
AUDIT round 2: stale=1 queued_jobs=1
AUDIT round 3: stale=1 queued_jobs=1

# dev
AUDIT round 1: stale=0 queued_jobs=0 rows=[(Some(...),"bound")]
```

Dev converges because it never writes the row. Each iteration rebuilds the ancestry closure and calls `materialize`, which now takes an advisory transaction lock and three statements over the base's facts (#797), so this is not a cheap spin — it is a base that never goes quiet.

The debounce in #757 coalesces runs; it cannot help here, because each run legitimately finds work it can never finish.

Smallest fix I can see: an orphaned binding is garbage, not a decision, so either `stale` joins the live signatures and stops reporting rows that no longer have one, or a run deletes bindings whose signature is gone. I would rather it were deleted — a negative decision about a signature that does not exist has nothing to say to anyone.

Your own `an_over_limit_stale_binding_is_not_falsely_rejected` asserts `jobs == 1`, which pins a narrower version of this same loop as expected. I would not cement that.

## Important, and not covered by any test here: the model is not told about the hierarchy

`keys_of(&p.domains)` still prints the declared domains, and the alignment prompt says nothing about subclasses. So a widened candidate reaches the model as

```
Item 5: phrase "located in" · subject class: manufacturer · object: place
Candidates:
- located_in · locatedIn · relation · domain: organization · range: place
```

and rule 1 tells it to choose a property "only when each of the statements under this signature states that property, by the definition as given". A careful model votes null on a domain that does not name the subject's class, and two null votes now write `none` — the structural negative your first half was built to stop. The scripted model in your fixture keys off the candidate text, so no test here can see it.

Either print the inherited path (`manufacturer → organization`) in the candidate line, or say in the prompt that a subclass satisfies a parent's domain. Half the fix is in the filter; the other half is in what the model is shown.

## The class-edit hole runs in both directions

The docs paragraph discloses one: a parent edit does not reopen a cached negative. The other is the direction this pull request is named for. Removing a class's parent edge makes an inherited binding structurally inadmissible, and nothing notices:

```
AUDIT after removing parent: status=bound
AUDIT after restoring parent: status=bound
AUDIT stale after restore: 0
```

`stale` reads `relation_types.updated_at` only; `entity_types` is never consulted, and ontology class routes enqueue `align_types`, which tail-enqueues `align_phrases` — the worker runs, `stale` selects nothing, the typed projection stays live. Your test reopens a negative by editing *an unrelated property*, which is honest but is not the class edit. Please say so in the docs paragraph, in both directions.

## Smaller

- **The new CI step can pass while running nothing.** A filter that matches zero tests exits 0: `test result: ok. 0 passed; 324 filtered out; REAL_EXIT=0`. Rename `worker_tests` or drop the `#[path]` mod and the step stays green. The neighbouring database step guards against this by counting passes into the step summary. Also, `--locked` is on this step and neither neighbour.
- `ancestry_is_directional_scoped_and_cycle_safe` asserts `len() == 4` and `== 2`, both of which include the node itself, because `class_ancestors` returns the reflexive closure. Nothing in the name, the test, or a doc comment says so — and `class_ancestors` has no doc comment, in a file where everything else does.
- `identical_class_keys_in_another_base_do_not_supply_ancestors` cannot fail: `parents` is keyed by UUID and built from this base only. It passes on dev and pays for a full second fixture to assert something structurally impossible.
- `structural_none_keeps_shared_person_support` has `assert!(!sources.contains(&first))` right after `assert_eq!(sources, vec![second])`. Dead.
- `a_saved_negative_is_materialized_after_a_failed_recompute` creates a function and a trigger on the shared `facts` table; the drop runs before the asserts, so a failed assert is safe, but a panic inside `f.run()` leaks both into a CI database where other suites are writing that table.
- Comments in `phrase_alignment.rs` are English; this repository writes them in Chinese. The module header still describes candidates as coming from declared domain and range only.
- The docs paragraph ends with "Do not rewrite decision timestamps or delete bindings to simulate reevaluation", which is an instruction to an implementer rather than design prose.

Adjacent and not yours: two agreeing votes naming a property key that `by_key` cannot resolve also land as `none`, indistinguishable from "nothing fits".

Maya-Kid:
Thanks for the detailed review and for reproducing these cases. I thought about this too narrowly: I focused on the local candidate filtering and negative-decision behavior without fully considering the lifecycle of signatures, stale bindings, and job convergence.

This seems more involved than I initially understood, especially with ontology changes and the context shown to the model interacting with that lifecycle. I need to step back and think it through more carefully before asking you to review another revision. Sorry for the extra review burden, and thank you for pointing this out.

Maya-Kid:
Withdrawing this implementation in favor of the architecture/design discussion in #807. The review showed that I had treated candidate filtering and retirement too locally; signature lifecycle, invalidation, and scheduling need a coherent design before another implementation. The code and regressions remain here as investigation evidence. Thank you for the detailed review.

PR #802, opened 2026-09-19
Count phrase signatures by statements rather than evidence rows

Maya-Kid:
A statement with multiple evidence rows inflates a phrase signature’s count and can occupy several of its three example slots. Select one stable quote per statement before grouping and sampling, preferring evidence with complete quote offsets. Evidence-free statements remain included and all evidence stays stored.

The PostgreSQL regression creates four open statements. On dev e894fa1, adding a second evidence row reports five statements; this fix keeps four and three distinct examples. It also checks that complete offsets outrank an earlier unlocated quote, multiple different located quotes are selected deterministically, Unicode offsets remain character-based, and invalidated statements are excluded.

Validation: the original implementation at 9ec6996 passed Linux workspace formatting, Clippy with warnings denied, build, and 920 tests (one existing live-network test ignored), plus GitHub CI. The test/comment supplement at e7cfec4 passed Linux formatting, the real-PostgreSQL regression with UTOPIA_TEST_REQUIRE_DB=1, and targeted Clippy with warnings denied. No production behavior changed in that supplement.

No migration, dependency or public API change. This corrects live signature reads; it does not refresh historical statement_count/examples already saved in bindings, change source freshness, or implement ontology proposals. UUID ordering is deterministic, not a relevance score; no scale benchmark is claimed.

Run with a dedicated PostgreSQL database and UTOPIA_TEST_REQUIRE_DB=1: `cargo test --locked -p utopia-store --test phrase_signature_evidence`.


WaylandYang (commented):
Verified. On an empty database the test passes; with only `phrase_bindings.rs` reverted:

```
Error: after 2 evidence rows, four statements counted as 5
```

That is the bug, and it is the same shape as one I ran into on the judge bench last week: a statement observed in two chunks is one statement and two evidence rows, and anything that counts the join counts it twice. Here it matters more than in a bench, because `statement_count` is what the workbench orders its suggestions by and what the aligner prints to the model as "N statements" — so the signature that looks most worth deciding may just be the one whose statements were seen most often.

The `LEFT JOIN LATERAL … ORDER BY (quote_start IS NULL OR quote_end IS NULL), chunk_id LIMIT 1` is the right shape: one row per statement, located quotes preferred over unlocated ones, and `chunk_id` as the tie-break so the three examples do not shuffle between runs. Keeping evidence-free statements in through the `LEFT` is right too — a statement with no evidence is still a statement, and dropping it would trade one miscount for another.

Two small things, neither blocking:

- The test is one function with five assertions in sequence, so the first failure hides the rest. The name says what the first one checks; the offset preference, the deterministic selection and the Unicode offsets are all in there too. Separate cases would tell you which invariant broke without reading the file.
- It calls `db::migrate`, so it cannot run against a database that already has the schema — I had to make an empty one. Two of your recent pull requests do this and two do not; the repository's convention is to assume a migrated database.

Nothing else. Merging.

Issue #800, opened 2026-09-19
A phrase decision recomputes the typed graph inside the request


WaylandYang:
From the review of #797, which is right to serialize materialization and does not cause this.

`decide_alignment_phrase` (`review_routes.rs:1280`) calls `materialize` inline, in the HTTP request. After #797 that call waits on a transaction-scoped advisory lock held for the whole recomputation of that base, and the wait has no bound — `acquire_timeout` covers checking out a connection, not waiting on a lock.

Measured on the branch:

- 500 statements: `first run 4.2545363s`, so a first bind on a large base is minutes of exclusive per-base lock.
- A waiting materializer pins a pool connection for the whole wait. Two waiters on a two-connection pool exhausted it: `third acquire after 3.0063752s => pool timed out while waiting for an open connection`.

The worker path is fine — `align_phrases` holds its own try-lock, so worker runs never queue. Review decisions have no such guard, so several people deciding phrases in one base queue on the lock, each holding one of 32 connections, each request hanging.

Two ways out, both leaning on machinery that already exists:

1. The handler enqueues rather than materializing, and the worker's end-of-run recheck picks up what arrived while it was running. That recheck is already in `phrase_alignment`, added in #757 for this exact shape.
2. The handler takes `pg_try_advisory_xact_lock` and answers "already recomputing" when it cannot have it. Instant and honest, and again relies on the recheck.

Either way the person's decision is saved immediately; what moves is when the typed rows appear. Worth checking whether the review page reads them back straight after deciding, since that is what would make the difference visible.

Related: #798, the same shape for a kind-word decision waiting on the aligner.

Issue #798, opened 2026-09-19
A review click can wait on the aligner with no bound


WaylandYang:
From the review of #792.

Deciding a kind word from the review page now writes the binding and its entity projection in one transaction, which is right: it is what stops an older agent run from applying its class after a person has decided. The transaction holds the `type_bindings` row from `decide` to commit, and inside it updates every entity in the base whose normalized `specific_type` matches that word. For a common word that is most of the base.

`decide_alignment_kind_word` is a synchronous HTTP handler, and the main pool sets no `lock_timeout` or `statement_timeout` — only `query_engine/postgres.rs` sets one, and that is for external databases. So a person's click can wait behind the aligner's entity update with no bound, and the only thing they see is a request that does not come back.

What it should do instead: set a `lock_timeout` on that handler's transaction and answer 409 when it expires, with a message saying the aligner is working on that word and to try again. A person who is told to wait a moment is in a different situation from a person watching a spinner.

Worth checking at the same time whether any other synchronous handler takes a lock that a background job can hold.

rollroyces:
Picking this up. Two things to settle before I write the patch:

1. **Scope of the fix.** Is it a single-handler fix in \`decide_and_apply\` plus the 409 response, or a survey of every synchronous handler that holds a lock a background job can take? The issue body hints at the survey ("Worth checking at the same time whether any other synchronous handler takes a lock that a background job can hold"), but the explicit ask is the kind-word handler. My read: do the kind-word fix as a focused PR, list the other candidates the survey finds in the PR description, and let you pick which become follow-ups. That way the safety win is a small surface and the survey is documented even if the follow-ups are not in this cut.

2. **The 409 message.** You wrote "the aligner is working on that word and to try again." I'd word it as: \`Retry-After: 5\n\` + a body that names the binding (so the UI can show "this kind word is being aligned, retry in a few seconds"). The agent path is \`align_phrases\` which is queued, not locking the row, so the lock holder is something else — an older decision applying, a recalc, etc. Should the message say what kind of work is holding the lock, or is "try again" sufficient?

3. **\`lock_timeout\` value.** Postgres default is 0 (wait forever). 5 s feels right for an interactive handler — long enough that a one-off slow query doesn't 409 spuriously, short enough that a person's click is never waiting on an aligner. Reasonable, or is the project using a different value elsewhere I should match?

[...]

PR #804, opened 2026-09-19
Discuss ontology input revisions for alignment

Maya-Kid:
A definition edit during type alignment can be hidden by the later reply timestamp (#795). This draft records a proposed snapshot/revision acceptance contract, the semantic writer inventory, and a reproducible correct-behavior test patch. The baseline test fails at the intended freshness assertion; the unchanged-input control passes.

Design discussion only: no production fix, migration or accepted ADR. The main decisions are type-only versus type-and-phrase scope, projection publication, endpoint-class generation and durable scheduling. Related to #795; does not close it.

Validation: documentation/link and patch-application checks. The attached reproduction was run on Linux with PostgreSQL and a scripted local endpoint; it is intentionally expected-red and is not enabled by this PR.


WaylandYang (commented):
The thinking here is wanted, and some of it is thinking I asked for: the acceptance protocol answers the question I left open in #775 ("in-flight model calls still race an edit, and that is input-version tracking, which is its own change"), and the lock-order section lands on the same place as #798 and #800 from the other side. Reading a snapshot in one short repeatable-read transaction, closing it before the model call, and comparing the revision again at acceptance is the right shape, and the insistence that a later reply timestamp is not freshness is exactly the mistake that is easy to make here.

What I do not want is where it lives.

## Three places for design is two too many

This repository already has two: `docs/design/*.md` says what is true now and carries a "Proposed and not built" list for what is not, and `docs/decisions/NNNN-*.md` says what was decided and why, with a Status line that gets updated as cuts land. `.roadmap-proposals/` arrived with #781 and I said the same thing about it there. `docs/design/proposals/` is now a third, and it arrives with three documents at once, all of them `Status: PROPOSED / DESIGN_REQUIRED, not accepted`.

A directory of drafts that are not accepted, not scheduled and not linked from anything is where documents go to rot. Six months from now a reader cannot tell which of them describes the system, which was abandoned, and which was quietly superseded — which is the whole reason the decision records carry a Status line.

Where each piece belongs:

- **The protocol itself** is a decision, not a draft. It changes the ontology contract and the data model, so it wants a decision record with a number, cuts, and thresholds — the form 0044 and 0045 use. If it is not ready to be decided, the argument belongs on the issue it answers.
- **The writer inventory** (`alignment-input-writers.md`) is a finding about the current code. That belongs on #795, or in the design page for alignment if it is durable.
- **The failing test, as a `.patch` file, should not be checked in at all.** 241 lines of diff stored as text is a test nobody runs and nobody updates; it is stale the first time `type_alignment.rs` moves. If it is worth having, land it as a real test marked `#[ignore]` with the reason in the attribute, so it compiles with every change and someone can run it deliberately. If it is not worth that, its value is the reproduction recipe, which is three sentences on the issue.

You say `#795 has no discussion response at this audit`, and that is fair — the answer to a quiet issue is to say more on the issue, which notifies people, rather than to commit a document that does not.

So: I would like this closed and the body posted on #795, with the test landed separately as an ignored test if you think it earns its keep. If you would rather it be a decision record, that is welcome too — pick a number, write the cuts, and I will read it against 0044.

The same goes for #805 and #806, and I have said so there rather than repeating the argument.

Maya-Kid:
Closing as requested in review. The findings and unresolved design questions are now on the existing issue: https://github.com/deeplethe/utopia/issues/795#issuecomment-5743377390. No new draft-document directory or unexecuted fixture is being proposed for merge. Thank you for clarifying the repository’s design and test conventions.

Issue #725, opened 2026-09-16
Review queues are organised by what a decision changes


WaylandYang:
Follows 0044 (#710) and 0045 (#724). A survey of the review and governance code on `dev` (about 12k lines: resolution, temporal, governance, review routes, type resolution, mappings, adjudication, pending; plus 2.2k in #699) against the three-layer design.

## Queues

Today's queues are grouped by where an item came from (pending, duplicates, conflicts, unconfirmed, lowconf, violations, defects, plus proposals on the ontology page and mappings). The new queues are grouped by what a person's decision changes. Agents handle every single item; people see what an agent could not settle or what one decision resolves at once.

| Queue | What arrives | The decision | Writes back to |
|---|---|---|---|
| Ontology proposals | object/link types and properties proposed by the ontology agent or drawn from the open graph's most frequent unbound signatures; duplicate properties; domain/range changes; definition edits; implication rules | approve, edit and approve, reject | ontology version; recompute of affected signatures; regression cases |
| Alignment | a signature (phrase + subject classes + object classes) the aligner cannot settle, and implication rules it proposes for a signature | property and direction, or none; accept or reject the rule | binding cache and rule set, reused across the base; "none" feeds a proposal |
| Time anchors | document date missing or conflicting; relative mention with no anchor; undefined calendar | the date or anchor | document time context; dependent valid times recomputed |
| Identity | pairs deterministic evidence cannot settle; merges the gate held | same, different, cannot-link | merge or split action; cannot-link list |
| Typed-fact errata | retractions and revisions the errata agent could not settle or the gate held | keep, retract, revise | fact invalidated plus its replacement, on the record axis |
| Cross-document conflicts | incompatible values of a functional property or incompatible times | which holds, both at different times, disputed | supersede, valid_to, dispute mark |
| Nods | memory facts awaiting the person (0015, 0043) | nod or not | fact takes effect |

Alignment is the operation that produces the typed graph from the open graph: bindings decided once per signature, and implication rules (a statement of this shape implies a fact of that property, with the value read from the object phrase) proposed by the agent, approved on the ontology workbench, and executed by code. The ontology workbench has three sources: suggestions from the open graph, an imported ontology file (packs, 0008), and online editing.

Ordering is by leverage (statements a decision resolves), then impact, then age.

## Cards

[...]

PR #805, opened 2026-09-19
Discuss evidence-backed ontology property proposals

Maya-Kid:
This draft defines a narrow path from reason-classified unbound signatures to evidence-backed relationship proposals and human adoption, reusing the existing workbench. It includes three-domain seed evaluation cases with explicit expected directions and reuse outcomes.

Design discussion only: no agent, migration or adoption endpoint is implemented. Decisions remain on proposal identity, evidence/version checks, rejection lifecycle and atomic adoption. The separate statement-count prerequisite is #802; this draft does not claim that fix implements the proposal flow. Related to #725 and ADR 0044.

Validation: documentation/link checks and JSONL parsing. Real-model quality and the new adoption lifecycle are NOT_RUN.


WaylandYang (commented):
The path you draw — a reason-classified unbound signature becomes an evidence-backed property proposal that a person adopts on the workbench — is the shape 0044 reserves for cut 5, and the seed cases with expected directions are the part I would have asked for. Naming #802 as the prerequisite rather than folding it in was the right call, and that one is merged.

The argument I made on #804 applies here: `docs/design/proposals/` is a third place for design in a repository that already has two, and a draft that is `not accepted` and linked from nothing is a document that rots. This one belongs on #725, or as a decision record with a number and cuts if it is ready to be decided.

One thing that is specific to this draft rather than to its location. `ontology-property-cases.jsonl` is a bench fixture without a bench: nothing reads it, nothing scores it, and its expected directions cannot fail. If the evaluation is real, it wants to live beside the thing that runs it — `scripts/bench/` has the corpora and the scoring for exactly this, and a case file there is checked by running it. As a file under docs it is a table of intentions.

Maya-Kid:
Closing as requested in review. The findings and unresolved design questions are now on the existing issue: https://github.com/deeplethe/utopia/issues/725#issuecomment-5743377511. No new draft-document directory or unexecuted fixture is being proposed for merge. Thank you for clarifying the repository’s design and test conventions.

Issue #507, opened 2026-09-08
A criterion stated in the corpus has nowhere to land


WaylandYang:
A well report says: *"当全烃大于 8 且解释结论为气测异常时，可判定为优秀井。"* — when total hydrocarbon is above 8 and the interpretation reads as a gas anomaly, the well counts as a good well.

That sentence is a **criterion**, and it is exactly the kind of thing [0021](../blob/dev/docs/decisions/0021-a-rule-reads-attributes-and-concludes-a-type.md) built a table for. Today it falls on the floor, and it falls necessarily rather than by oversight.

## Why it cannot land

- **`facts.subject_id` is `NOT NULL`.** Every fact hangs on an entity. This sentence has no subject — it is about a *class* of wells, not about any well. There is no entity to hang it on.
- **Extraction emits triples.** A universally quantified conditional does not fit in three slots.

So the likely outcome is that it is dropped. The worse outcome is that it is attached to whichever well happens to be nearby in the context, turning "here is how to judge a well" into the asserted fact "W-1 is a good well" — a claim the document never made.

## Where it should land

**`attribute_rules` is already the right shape.** It holds "an entity of this class, whose own attributes satisfy these conditions, concludes this" — which is what the sentence says, slot for slot. The class, the conditions, the conclusion are all there.

**What is missing is not the shape. It is the provenance.** `attribute_rules` has no `chunk_id`, no `document_id`, no evidence of any kind: it assumes a rule is something a person typed into the ontology page. The one in the corpus has a source — page N of a particular report.

## The line 0002 drew, and the half of it that does not apply

[0002](../blob/dev/docs/decisions/0002-reasoning-engine.md) decision 5 says the reasoner's criteria are written by people and rules the model out. That ban is really two claims, and only one of them is at stake here:

- **The model invents a criterion** — refused, and it stays refused. One authored criterion re-decides the whole base, and nobody can say where it came from.
- **The model reports that a document states a criterion** — this is a *fact about a document*. It has a sentence, a page, a filename, exactly like every other thing extraction produces.

The line between them is **whether it has a source**, and a source is the thing this system has the most of.

That is also precisely the distinction [0015](../blob/dev/docs/decisions/0015-recording-a-sentence-is-not-asserting-a-fact.md) drew for facts: recording a sentence is not asserting a fact, and `pending_facts` is where the difference lives. **A criterion found in the corpus should be to a rule what a pending fact is to a fact.**

0021's open question already sketched this shape before there was a reason to build it:

[...]

PR #806, opened 2026-09-19
Discuss sourced criteria as reviewable rule proposals

Maya-Kid:
A document criterion needs provenance and human review before it becomes an executable rule (#507). This draft proposes a narrow constant-comparison slice that reuses the existing engine, with narrative, exception, unit and threshold cases.

Design discussion only: no rule proposal storage, extraction or adoption is implemented. The key decision is whether an approved rule remains under human authority after its source changes or is suspended as a source-dependent projection. The draft also identifies atomic adoption and a nonpersisting preview as required work. Related to #507; does not close it.

Validation: documentation/link checks and JSONL parsing. Existing engine tests passed in the prior Linux workspace run; this is not validation of the proposed sourced-rule feature.


WaylandYang (commented):
The question you put at the centre is the right one, and it is the one I would want answered before any code: when the document a rule came from changes, does the approved rule stay under human authority or get suspended as a source-dependent projection. That is a decision, and 0044's treatment of a person's decision — never overwritten by the agent — is the precedent it has to argue with.

Same as #804 and #805 on where it lives: `docs/design/proposals/` is a third place for design, and a draft that is not accepted and not linked is a document that rots. This belongs on #507, or as a numbered decision record if the question is ready to be settled.

Specific to this one: `sourced-rule-cases.jsonl` has the same problem as #805's case file. Narrative, exception, unit and threshold cases are worth having, but under docs nothing runs them, so nothing can tell you when one stops holding. The rule engine has tests; that is where a case file earns its keep.

Maya-Kid:
Closing as requested in review. The findings and unresolved design questions are now on the existing issue: https://github.com/deeplethe/utopia/issues/507#issuecomment-5743377629. No new draft-document directory or unexecuted fixture is being proposed for merge. Thank you for clarifying the repository’s design and test conventions.

PR #815, opened 2026-09-20
Render computed rule conclusions from their stored expressions

Maya-Kid:
A computed rule stores its expression in `conclude_expr` and has no constant `conclude_value`. `list_rules` currently describes `revenue - cost` as `margin = null`.

Render the stored expression instead. Reuse the existing expression validation and depth limit, resolve attribute names in one KB-scoped batch query, and preserve the tree with parentheses. Labels retain their keys when different, so duplicate labels remain distinguishable. Unreadable expressions or missing references explicitly say `(expression unavailable)`.

Validation on Linux / Rust 1.98.1 and isolated PostgreSQL:

- The authenticated MCP regression fails on unchanged `75951c7` (`margin = null`) and passes with this patch.
- Covers subtraction, nested subtraction/division, zero and negative constants, Chinese/duplicate labels, cross-KB references, malformed expressions, and unchanged typing/constant conclusions. Rule/condition descriptions, derived rows and scoped job snapshots remain unchanged by reads.
- `cargo test --locked -p utopia-server api::`: 80 passed with the database required.
- A separate local compatibility branch applies #808 cleanly; all 13 authenticated MCP tests pass together, including both regression tests.
- Formatting, workspace Clippy with warnings denied, workspace tests and workspace build pass with the database and Poppler required. The existing live-HTTPS acceptance test remains ignored.
- The unchanged frontend tree shares the earlier same-host validation: frozen-lockfile install, 91 tests, and production build passed.

This only describes definitions; it does not evaluate expressions or change materialization, migrations, jobs, or historical data. The condition-group fix in #808 is independent and is not included in this diff.


WaylandYang:
Merging this. Describing `margin = null` for a computed rule was misleading in the one direction that matters: an agent reading it would conclude the rule concludes nothing.

Two small notes, neither blocking.

The renderer parenthesises every node, so a two-term expression arrives as `margin = (revenue - cost)`. The outer pair carries no information; dropping it at the top level and keeping it for nested nodes would read closer to how someone would write the formula. Your tests pin the current form, so it is a change for you to make if you agree rather than one to bolt on at merge time.

`format!("{label} [{key}]")` when the label differs from the key is the right call for disambiguation, but it doubles the width of every leaf in a deep tree. If expressions get long in practice, the key is only needed when two leaves in the same expression share a label.

PR #816, opened 2026-09-20
Preserve validity intervals in rule-match descriptions

Maya-Kid:
`rule_matches` returns accepted derived rows, including historical intervals, but its text drops both validity bounds and its step card calls the row count “entities”. A real materialization of two disjoint intervals for Alice produces two indistinguishable lines.

Include each interval, formatting its endpoints with the existing precision-aware `time_text::world`. Fetch the two existing precision columns alongside the existing bounds. When at least one bound is known, a missing opposite bound says `unknown start`/`unknown end`. Rows with neither bound omit the validity bracket; they do not assert current validity. Count and truncation text now say “matches”. Filtering, ordering and row pagination are unchanged.

Validation on Linux / Rust 1.98.1 and isolated PostgreSQL:

- The authenticated MCP regression fails on unchanged `75951c7`, after the real materializer produces two historical rows for one entity. The patch passes.
- Covers typing and attribute conclusions, year/month/day precision, unrounded microseconds when precision is absent, start-only/end-only intervals, omitted brackets for fully unbounded rows, pagination and step counts, unchanged premises, invalidated-row exclusion and KB scoping. Reads leave rule/condition, derived-row and job snapshots unchanged.
- `cargo test --locked -p utopia-server api::`: 85 passed with the database required, including existing empty/error/authorization controls.
- Formatting, workspace Clippy with warnings denied, workspace tests and workspace build pass with the database and Poppler required. The existing live-HTTPS acceptance test remains ignored.
- The unchanged frontend tree shares the earlier same-host validation: frozen-lockfile install, 91 tests, and production build passed.

The shared store response gains two precision fields; the existing UI ignores these additive fields. This PR changes the agent tool description, not the UI's own date rendering. It adds no migrations, does not change world-time filtering or materialization, and does not backfill data.


WaylandYang:
Merging this — the bug is real, `rule_matches` was returning two indistinguishable lines for two disjoint intervals, and the precision-aware formatter is the right tool. One observation to weigh, not a change request.

`[validity: unknown start → unknown end]` now appears on every row whose bounds are both missing, and that is not a rare row. On the four SEC filings I measured this week, 546 of 885 open statements had neither a start nor an end, and another 274 had only an end. A base built from prose rather than from tables will be mostly that shape, so most lines in a `rule_matches` answer will carry a bracket that says nothing twice.

Your reasoning for printing it is sound and I would not silently drop it: a missing bound must never read as "holds now". But there is a third option between silence and `unknown → unknown`: print the bracket only when at least one bound is known, and keep the fully unbounded rows bare. Nothing is then asserted about them, and the rows that do carry a date stand out instead of being buried in boilerplate.

Your call. If you keep it as is, the one thing I would change is the arrow spacing, since `unknown start → unknown end` reads as a range between two things that exist.

WaylandYang:
Your branch went stale while the other nine landed: #808 and #815 both added tests next to yours in `mcp_tests.rs`, and git could not tell three new test functions apart from one edited one.

I rebuilt it on top of current `dev` rather than hand-merging three hundred lines of conflict markers: your changes to `tools.rs` and `business_rules.rs` replayed cleanly with a three-way apply, and `rule_matches_keep_materialized_intervals_and_count_rows` went back in whole at the end of the file. Nothing of yours was dropped or rewritten. `cargo fmt --all --check` and workspace clippy are clean, and the test passes against a database on the new base:

```
test api::mcp::tests::rule_matches_keep_materialized_intervals_and_count_rows ... ok
```

Force-pushed to `codex/rule-match-validity`, so pull before you continue on that branch.

The note I left earlier about `unknown start → unknown end` on unbounded rows still stands as a question for you, separate from this.

PR #834, opened 2026-09-20
feat(export): serialize all supported ledger surfaces

ansonnmm:
**PROBLEM**

Rules, attribute rules, document versions, chunks and evidence are read from the export snapshot but never serialized — their facts' `prov:wasGeneratedBy`, `prov:wasDerivedFrom` and locator references dangle in the finished file. Newer ledger surfaces are absent too: class/relation `updatedAt`, entity descriptions, document reader/time context, chunk origin metadata, evidence quote offsets, open-statement layer/phrase/qualifiers/time mentions, typed-fact `fromStatement` provenance, `validFromGrade`, and `evidenceOrigin`.

The serializer was never extended as the ledger schema grew, so the export contract silently covers only a subset of what the read model returns — a file that claims to be the ledger but omits whole families and leaves references pointing at nodes that were never written.

**FIX**

Serialize every family in snapshot order: vocabulary (classes, relations), rules and attribute rules with their conditions and expression-predicate references, documents and document versions, chunks, entities, facts with qualifiers / time mentions / lifecycle anchors, evidence, and derivations with ordered premises. Emission fails closed — an unresolvable reference aborts the export rather than minting an IRI for a node that was never written.

**REGRESSION EVIDENCE**

New in-memory unit tests cover each emission family in both Turtle and JSON-LD: rule conclusions and conditions, expression predicates, document-version locators, chunk/evidence links, open-statement ledgers, qualifiers (literal and entity), lifecycle marks, imported vocabularies, and fail-closed paths (unresolvable qualifier type, merged entity reference, foreign IRI minting). 32/32 `rdf::` tests pass.

**COMPATIBILITY RISK**

Additive only: exports grow by the previously missing families; no existing triple is removed or renamed.

---

Stacked on #832 → #833 — this PR's diff includes those commits until they merge.


ansonnmm:

PR #835, opened 2026-09-20
test(export): account for every emitted RDF quad

ansonnmm:
**PROBLEM**

The export serializer's unit tests assert individual triples. Nothing verifies the output as a complete set — a silently dropped family, a duplicated quad, or a triple landing in a named graph would pass every existing test.

Per-triple assertions cannot catch absence or duplication. A serializer contract that claims "the whole ledger" needs set-level accounting, not sampling.

**FIX**

Test-only change (`#[cfg(test)]`). An in-repo oracle independently derives the expected quad set from a synthetic fixture — without calling any `emit_*` function, including its own literal/leaf construction — and requires exact equality in both Turtle and JSON-LD, all in the default graph, with `emitted == distinct`. A declared predicate matrix registers every conditional cell per node kind (including cells only reachable through synthetic rows), a coverage test requires every declared cell exercised in both its present and absent branches, and a mutation test requires unaccounted / moved / dropped / named-graph terms to break the accounting. New ledger surfaces must register their predicates in the matrix, so the accounting stays complete as the contract grows.

**REGRESSION EVIDENCE**

This change is the regression capability: `the_whole_export_is_accounted_for`, `the_accounting_rejects_unaccounted_terms`, and `the_fixture_exercises_every_declared_conditional_cell` all pass — 35/35 `rdf::` tests on the branch.

**COMPATIBILITY RISK**

None at runtime — `cfg(test)` only. The cost is maintenance: ~2.4k lines of test code that must be updated when the serializer contract grows; the declared matrix makes that required update explicit rather than silent.

---

Stacked on #832 → #833 → #834 — this PR's diff includes those commits until they merge.


ansonnmm:

PR #836, opened 2026-09-20
docs(export): document the RDF export contract

ansonnmm:
**PROBLEM**

The documented RDF export contract describes entities, facts and documents — but not the serialized rule, version, chunk and evidence families, the reference-resolution guarantees, or the single-snapshot and fail-closed semantics the route now provides. It also says nothing about the ledger surfaces newer migrations added.

The public contract documentation lagged the implementation: consumers cannot rely on guarantees that were never written down, and the documented subset understates what an export actually contains.

**FIX**

Document the whole emitted contract in `web/src/docs/mcp.md`: open statements and their qualifier/time-mention nodes, typed-fact `fromStatement` provenance, validity grades, evidence origin/offsets, chunk producer metadata, document reading context, vocabulary `updatedAt`, the single-snapshot read and fail-closed semantics, and the explicit non-claims.

**REGRESSION EVIDENCE**

Documentation only; the documented behavior is covered by the tests in the underlying PRs.

**COMPATIBILITY RISK**

None — docs only. The doc declares the export a compatibility boundary: new predicates are additive, while omission or renaming becomes contract-sensitive.

---

Stacked on #832 → #833 → #834 — this PR's diff includes those commits until they merge.


ansonnmm:

PR #839, opened 2026-09-20
design: validate a conservative expression declaration policy

Maya-Kid:
ADR 0032 asks for declaration checks and a usable nested picker, but the standalone model did not exercise the real interface or a real knowledge base. This revision puts the decision in ADR 0049 and replaces the disposable picker with a reusable controlled React editor at the unlisted `/kb/$kbId/expression-draft` route.

The route uses the existing authenticated ontology/rules APIs, SearchSelect/Dropdown/Input controls, design tokens and both language packs. It only previews an unsaved structured draft. It does not enable unit validation, save rules, change B1 metadata editing or change dependency analysis. Unsupported expressions stay unconverted; incomplete constants and missing references cannot become a zero or a default attribute.

The policy module, Node model tests, PostgreSQL probe and standalone HTML were archived outside the repository and removed from the PR tree. ADR 0049 preserves their historical findings, exact source revision and limitations. Both index tables are updated; 0032 has a dated note beside its unit-check discussion.

Validation at `64206696689f1901f5cfd3420553a26d70b8de3a` on latest dev `69560db5969fae8def8c4e84b7572d841e1dca57`:
- `pnpm guard`, `pnpm typecheck`, `pnpm test` (119 passed), and `pnpm build` passed on Linux, Node 22.23.2 / pnpm 10.2.1.
- Headless Chrome against the real Linux backend read API-created bases with 30, 300 and 1000 attributes, including duplicate/long bilingual labels, mixed datatypes and absent units. Search reached attribute 0999 by key; keyboard selection/focus, margin, right-nested subtraction/division, internal edits, unfinished numbers, depth-four cap, 390px width, empty bases and SPA KB reset passed. The draft UI sent no write requests.
- Loading, Chinese UI and actual restricted-KB authorization passed. Read/retry and unknown-shape checks additionally used explicitly injected browser response faults.

Four levels remain a long vertical form on a narrow screen; these checks are not a human usability score. Exact `1`, the known-unit allowlist and first-cut ratio policy remain pending. There is no B2 persistence flow or production unit validator. Draft status is intentional.

Refs #488


WaylandYang:

Specific to this one: the unit policy table and the `FOR SHARE` versus `FOR KEY SHARE` transaction boundary are both answers to open questions 0032 already carries ("Units and datatypes have to be checked when the expression is written, and today nothing checks them"). That makes them a revision of 0032 with a record of their own, not an appendix.

WaylandYang:
Adding one more, on `scripts/prototypes/expressions/index.html` specifically: **the picker experiment should not be a standalone page.**

The reason for isolating it is understood and half right — `policy.mjs` is the *proposed* policy, not the one in force, and 0032's picker is still an unbuilt cut, so there is nothing in `RulesPanel` to extend and no case for putting an unapproved policy on a production code path before the decision.

But the isolation costs more than it saves, because of what the page is being asked to answer. 0032's open question is "**How deep before the picker loses**", with the record's own example being that `(revenue − cost) / revenue` as nested dropdowns is worse than typing it. That is a question about whether the thing is bearable to use — and this page cannot answer it:

- It runs on three hardcoded attributes (`revenue`, `cost`, `ratio`). A picker gets bad exactly when a real base puts hundreds of attributes behind each `<select>`, which is the case the experiment never sees.
- It makes no API call at all, and "save" stringifies into a `<pre>`.
- It reimplements every control with `document.createElement`, with inline hex colours (`#111827`, `#64748b`) and `font: 16px system-ui`. Living outside `web/` also puts it outside `pnpm guard`, so it would be the one piece of interface in the repository that no style rule reaches — while the shadcn migration and the token bridge are the standing direction for everything else.

So what it demonstrates is the interaction *shape* in a mock, and the shape was not in doubt. Put it in `web/` behind a route or a flag, built from the real components and reading a real base's attributes, and the depth question gets a transferable answer — and nothing has to be thrown away when the real picker lands. As a standalone file its fate is to become a second implementation that rots the day cut B2 ships.

The policy module and its tests are a different matter; `policy.mjs` being separate from the production validator is fine and is not what this note is about.

WaylandYang:

Applied to this PR:

- `policy.mjs` and `policy.test.mjs` — the executable spec described above. Whichever way the decision goes, it has no future: approved, the production validator implements it and this drifts; rejected, it is dead. The unit table and the operand rules are the part worth keeping, in the record.
- `test_declaration_lock.py` — what it establishes is one sentence ("`FOR SHARE` blocks a concurrent declaration `UPDATE`; `FOR KEY SHARE` does not"), plus the transaction-boundary proposal. Both belong in the record. The script is the means, not the finding.
- `index.html` — covered separately above: it should be a route or flag inside `web/`, built from the real components against a real base's attributes, because the question 0032 actually asks is about depth in a real picker.

WaylandYang:

WaylandYang (approved):
Picking this up after the ready-for-review flip — the rework matches what was asked
for on the 20th/21st, and nothing since then changes the shape: two index-sync
commits to keep 0049's rows next to 0050/0051 as they landed.

Verified on a merge of this branch onto current `dev` (after #833/#845/#860/#864):
merges cleanly, both README index rows land in order between 0048 and 0050,
`pnpm guard` 99 files compliant, `pnpm test` 132 passed (the three new
`expressionDraft` cases are collected by vitest, not just present), `pnpm build`
including type-check passes. No backend, no migration, both language packs updated.

What lands is a **proposed** record, same as 0047/0048/0050 — the exact spelling
of `1`, the known-unit allowlist and whether ratios are in the first cut stay open
in the record's own status line for whoever settles them. The unlisted route is
read-only, non-persisting, and says so on screen; it stays unlisted until 0049 is
approved and the implementation folds it into RulesPanel, per the record's own
plan.

LGTM.

Issue #530, opened 2026-09-09
Actions: a declared call the base can make


WaylandYang:
A base can conclude (0021, 0032) but cannot act: nothing holds a call it may make, a typed parameter, or a record of what it sent to another system. Record 0034 designs the piece: an **action** is a declared HTTP call (method, URL, headers, a sealed auth block, a JSON body template, scalar parameters with bounds or a value set), registered once for the deployment and **granted** to the bases that may call it, with every run a row read by a deployment-wide log and each base's own.

Cuts:

1. Capability: schema, `utopia-store::actions`, the runner (render, send through `http_fetch`, record), routes, tests.
2. Administration: the registry and the action log; the base's Actions tab with what it was granted, a Run dialog and its log.
3. Rules fire actions: its own record.

PR #840, opened 2026-09-20
design: specify action dispatch identity and uncertain outcomes

Maya-Kid:
A synchronous run-then-record cannot distinguish an unsent action from a remote side effect whose response was lost. ADR 0050 records the proposed execution identity, revision-bound preview, original-flow dispatch grant and uncertain outcome, including why retries and recovery takeover are unsafe without another contract.

This revision is documentation-only. The Python protocol model, tests and requirements file were archived outside the repository and removed from the PR tree. Both decision indexes now contain 0050, and 0034 has a dated note at its synchronous-run boundary. No production schema, sender, route or recovery behavior is introduced.

Validation at `8f3d189d46478f437df57b4a823556dfc493dcbf` against dev `69560db5969fae8def8c4e84b7572d841e1dca57`: diff scope, two index entries, relative links, contextual revision note, proposed status and absence of executable scaffolding all passed. No fresh Rust/sender test claim is made for this documentation-only revision. The 19 model tests and mutation findings in the record are explicitly historical, tied to the original model revision.

Request identity and revision binding, no automatic retry, no redirects and no recovery takeover remain proposals for maintainer review. HTTP observation remains distinct from business completion; this does not promise external exactly-once execution.

Refs #530


WaylandYang:

WaylandYang:

Applied to this PR: `protocol.py` and `test_protocol.py` are 422 lines of Python sketching a dispatch protocol for a Rust system, and the file's own docstring says it is not Utopia's action implementation. When 0034 is built it will be built in Rust, and this becomes a design document that happens to execute. `requirements.txt` would also open a Python dependency line in a Rust and TypeScript repository, which is a bigger commitment than this experiment needs.

The identity and fault-boundary model is the valuable part, and it reads as decision material — put it in the record.

WaylandYang:

PR #841, opened 2026-09-20
design: validate durable delivery of human phrase decisions

### LINKED ISSUE #800: A phrase decision recomputes the typed graph inside the request

Maya-Kid:
The durable-delivery reasoning now lives in ADR 0051, indexed in both tables, with a dated contextual note in 0044. The one-off script README is removed; the real Rust regressions stay.

The unused public `try_materialize` entry point is removed. Production `materialize` still calls private `materialize_in_tx`, and production `phrase_bindings::decide` calls `decide_on`. Validation still precedes pool acquisition on the existing decide path. The integration target is now `human_phrase_materialization_delivery`; normal recomputation and process-exit cases use the existing public materializer.

Busy orchestration lives only in `materialize_delivery_tests.rs` under `cfg(test)`. It acquires the try-lock, calls the same real private body after success, and uses real Deferred handling on Busy. The test observes production `materialize` blocked in PostgreSQL by that lock, releases it, and verifies completion. The old-to-new assertion map retains rollback, pool release, same-job deferral, late arrivals, reverse/duplicate processing, actual worker startup, bounded deferral, scoped requeue, cost and all three OS process-kill windows.

Linux validation at `7e49b7fd6d555727f169a77d36ab12367bf97003` against dev `69560db5969fae8def8c4e84b7572d841e1dca57`:
- `cargo fmt --all --check`, `cargo clippy --locked --workspace --all-targets -- -D warnings`, `cargo test --locked --workspace -- --test-threads=1`, and `cargo build --locked --workspace` passed; database and PDF requirements enabled. Workspace: 959 passed, 0 failed, 5 ignored (existing live HTTPS plus the opt-in entries below).
- Dedicated idle PostgreSQL 16.15 database: both integration parents passed and actually killed/waited for three children (before commit, after acceptance, after projection before ack). The private busy-lock parent passed separately.
- Splitting decision/enqueue transactions, treating Busy as done, and requiring a model each caused the intended runtime assertion failure. Restored code passed again.

Opt-in commands are in CONTRIBUTING and the test module headers. No production job kind, HTTP 202/status read, Review UI or multi-instance recovery contract is introduced. Partial-projection process kill and failed ack persistence remain future production acceptance; this does not close #800. The proposed contract remains Draft.

Refs #800


WaylandYang:

Separately, and this one does block: **`try_materialize` is public, unused, and in a production crate.** Your own addendum says unused prototype helpers alone do not resolve #800 — which is exactly the problem. A `pub async fn` in `utopia-store` reads as wired to the next person who greps for it, and `dead_code` will not warn on it because it is public. Either land it together with the caller that uses it, or keep it out until that caller exists.

The `materialize_in_tx` extraction and `phrase_bindings::decide_on` are a different matter and are fine — those are enabling refactors with real callers, and pulling the body out of `materialize` so a transaction can be owned by the caller is a good change on its own.

WaylandYang:

Applied to this PR, with one carve-out:

- `scripts/prototypes/phrase-delivery/README.md` — the decisions requested go in the record.
- **`crates/utopia-store/tests/phrase_delivery_prototype.rs` should stay.** It is not a prototype: it is a Rust test in the real crate, exercising the real schema through the real store functions, and it records genuine database, process-exit and mutation evidence. Please just rename it to say what it tests rather than that it is a prototype — the word is what makes it look disposable.
- `try_materialize` still needs an answer, separately from all of this: it is public, unused, and `dead_code` will not warn on it.

WaylandYang:

PR #847, opened 2026-09-21
fix(store): degenerate record-axis predicates to the single-sided for "now"

plpycoin:
…m for "now"

The record-axis helpers always emitted the two-sided replay form (invalidated_at IS NULL OR invalidated_at > coalesce($n, now())). The OR-branch stops the planner from proving invalidated_at IS NULL, so the partial indexes (facts_live_subject_idx / facts_live_object_idx and friends) become unusable: graph overview's per-entity degree subquery degrades to a full seq scan per entity. On a KB with 9.3k entities x 95k facts the node query alone took 61.6s (EXPLAIN ANALYZE: Seq Scan on facts, loops=7665); with the degenerate predicate it is 4.3s via BitmapOr over the two partial indexes.

Follow owner_at's precedent: every predicate helper now takes Option<usize> and emits the plain current-state form (invalidated_at IS NULL, status = 'open', deleted_at IS NULL, ...) when no as_of is requested — the module docs always promised this degeneration but the implementation never did it. Call sites pass as_of.map(|_| N); the three hand-rolled Some/None workarounds (search_entities, same_name_peers, blocked_for_entity) collapse into a single call since entity_visible_at(None) is equivalent to merged_into IS NULL, as its doc comment already stated.

graph::overview's node query additionally passed Some(3) for the degree subquery unconditionally — that was the hot path. It now follows as_of.


WaylandYang:
Reproduced this independently, and the mechanism is exactly as you describe.

Same query shape on a local base — `utopia_fix`, 7,984 entities and 13,877 facts, so roughly seven times fewer facts than yours — over 500 entities, `EXPLAIN (ANALYZE, BUFFERS)`:

| | two-sided | degenerate |
|---|---|---|
| total | 146.8 ms | **9.6 ms** |
| shared buffers hit | 37,597 | **2,982** |
| heap blocks | 33,500 | **976** |
| rows removed by filter | 1,849 per loop | 0 |

The plans say why. The two-sided form recheck is

```
((kb_id = e.kb_id) AND (invalidated_at IS NULL)) OR ((kb_id = e.kb_id) AND (invalidated_at > COALESCE(now())))
```

— keyed on `kb_id` alone, with `subject_id`/`object_id` pushed into a heap filter. The degenerate form gets the BitmapOr over `facts_live_subject_idx` and `facts_live_object_idx` that those indexes exist for. 15x here against your 14x on a larger base is consistent, and the smaller absolute numbers are just the smaller table.

What convinced me on correctness is not the benchmark though — it is that **three call sites had already hand-rolled this degeneration**:

```rust
let visible = match as_of {
    Some(_) => crate::record_axis::entity_visible_at("e", 3),
    None => "e.merged_into IS NULL".to_string(),
};
```

So the semantics you are moving into the helper are the semantics three readers independently decided were right, and `node_sql(Some(3), ...)` was the one that forgot. Consolidating them is the change; the hot path was collateral. The module doc already promised the degeneration, so this makes the implementation match what it always claimed.

One thing worth stating out loud in the module doc while you are in there: the degeneration is sound because no row carries a future `invalidated_at` / `deleted_at` / `decided_at` / `resolved_at` — retraction always stamps `now()`. That invariant is asserted in prose and enforced by nothing. Before this change one call site was accidentally robust to a violation of it; now none are. I am not asking for a CHECK constraint — just for the assumption to be named where the next reader will meet it, since it is now load-bearing in four predicates rather than one.

**Blocking**: DCO is failing — the commit has no `Signed-off-by`. Please amend with `-s` and force-push, and I will approve the CI run (fork PRs need that here after every push).

Also worth answering in the description, per the convention in `docs/decisions/README.md`: which record this touches. My reading is that it implements what 0019's machinery always documented rather than changing any decision, so no status line needs updating — but the question should be answered rather than left open.

plpycoin:
## Decision record

  Implements [0019](docs/decisions/0019-the-second-clock-can-be-rewound.md)'s `held_at(T)` machinery as documented ("`None` means now" — the record's module doc in `record_axis.rs` always promised the
  degeneration; this makes the implementation match it). It overturns no decision: the two-sided form was an implementation gap, not a recorded choice. No status line needs updating.

  Review 评论回复：

  > Done both. The invariant is now named in the module doc: no row carries a future invalidated_at / deleted_at / decided_at / resolved_at — retraction always stamps now() — with the explicit note that nothing
 enforces it but this paragraph, and that post-degeneration all four predicates rely on it. Commit amended with -s and force-pushed (13a5d4b); DCO should be green. On the record question: this implements 0019
as documented and overturns nothing, so no status line changes — the answer is now in the PR description as well.

WaylandYang:
Reviewed and reproduced. Thank you — this is a real find, and the write-up of the invariant it leans on is what made it reviewable.

**The mechanism checks out independently.** On a KB here (1,278 entities / 4,465 facts) with `enable_seqscan = off`, so the question is only whether the partial indexes are *reachable*:

| predicate | index chosen | rows per loop | time |
| --- | --- | --- | --- |
| two-sided (`IS NULL OR > now()`) | `facts_live_time_idx` + `facts_invalidated_idx` | 9,877 | 1079 ms |
| degenerate (`IS NULL`) | `BitmapOr(facts_live_subject_idx, facts_live_object_idx)` | 3 + 1 | 133 ms |

Same shape you reported, smaller data. The OR-branch really does stop the planner proving `invalidated_at IS NULL`, and `facts_live_*_idx` becomes unreachable rather than merely unattractive.

**The invariant holds today.** Every write to `invalidated_at`, `deleted_at`, `superseded_at`, `decided_at`, `resolved_at` and `reverted_at` in `utopia-store` and in the migrations is either `now()`, `NULL`, or copied from a column that is already in the past. No call site parameterises one. So `IS NULL` and `IS NULL OR > now()` agree, and putting that in the module docs is the right place for it given nothing enforces it.

**Two things pushed onto the branch** (9fd71a1f):

1. `entity_detail` still passed `node_sql(Some(3), …)` — the last literal in the file, and the entity panel is opened on every node click. Its blast radius is one seq scan rather than 7,665, so it is not the hot path, but leaving it contradicts the point the PR makes.
2. Three unit tests in `record_axis`. The degeneration is a performance property with no other guard; a plain string assertion is enough to stop a later refactor quietly restoring the second branch. They cover all seven `None` predicates, the replay form, and `entity_visible_at`.

**One cost the description does not mention.** `entity_visible_at(None)` is not only a collapse of three hand-rolled branches — for `search_entities`, `same_name_peers` and `blocked_for_entity` it replaces the plain column test `e.merged_into IS NULL` with `NOT EXISTS (… entity_merges …)`. Measured here: 0.93 ms → 2.55 ms over 1,278 entities and 376 merges, same row count. It plans as a hash anti-join, so it grows with the sum of the two tables rather than their product, and it buys a single call site. Acceptable, and worth knowing about.

Replay is unaffected: the 18 record-axis rewind tests pass, as do the 27 tests that touch `entity_detail`. Merging once CI is green.

PR #855, opened 2026-09-21
Make older chat conversations reachable through existing pagination

Maya-Kid:

Based on `dev@474b904dc3e8014d42404a3c067af9906036a953`; current head `d01cc796b9ad2e3911fd56cb8133763c2b121b8c`.

Merge order follows the maintainer’s request: **#854 → #856 → #855**. This branch is now stacked on #856 (and therefore #854), with pagination replayed on top and import/i18n conflicts resolved. Its dev diff includes those prerequisites until they merge. The stacked frontend separately passed **116 module tests, 27 browser subtests**, and build/typecheck/guard. Maintainer visual review is still required.

---

The conversation rail only requests the first 30 records even though the existing list endpoint already provides `limit`, `offset` and `total`. Older conversations remain available by URL or search, but cannot be reached by browsing the rail.

Use the existing infinite-query support to load bounded 30-record pages, scoped by knowledge base and search term. Keep loaded rows on next-page failure and retry the same offset. Deduplicate by conversation ID, and let list invalidation refetch the loaded page range when updated conversations change the ordering. No API, database schema or dependency changes are introduced.

Validation on `dev@ea0557b`, fix `9ed91f7`:
- The 65-conversation browser regression fails on the original Chat component because there is no way to request the next page.
- All 11 real-component browser subtests pass: 65 records; 0/30/31/60 boundaries; failure/retry; stale pages after search or KB changes; duplicate IDs with identical titles; multi-page search; and refetching loaded pages after reordering. These tests use controlled HTTP responses with the real router/query client/Chat component.
- Frontend 116 module tests and build/typecheck/guard pass.
- Chrome against the actual Linux server and PostgreSQL browsed 65 synthetic persisted conversations through the real list API and reached all 65.
- Separately, integration tree `f7bcd0878bf490d6efb122b71d6f13d379df4709` with #845 and the other scoped Chat fixes passes Rust fmt, strict workspace Clippy, 1,013 tests (one pre-existing external-HTTPS RSS test ignored), workspace build, 129 frontend module tests, 27 browser subtests and frontend build.

The browser test uses optional Playwright/Chromium, matching the existing browser-test pattern. Run `node --test tests/chat-pagination.test.mjs` from `web`; `CHAT_PLAYWRIGHT_PATH` and `CHAT_CHROMIUM_PATH` may specify local installations.

Offset pagination still does not provide a snapshot across concurrent updates. Deduplication prevents repeated rows; a later invalidation refreshes the loaded range. This patch does not claim stable traversal under arbitrary concurrent writes.


WaylandYang:
Merge-order note for the three Chat view PRs, from trial merges onto `79caca04`:

```
OK        #854
CONFLICT  #855   web/src/pages/Chat.tsx  web/src/i18n/en.ts  web/src/i18n/zh.ts
OK        #856   (clean on top of #854, as its stated dependency implies)
```

So the order is **#854, then #856, then this one**, and this is the one that needs the rebase. Nothing surprising — all three edit `Chat.tsx` and add adjacent i18n keys.

Two notes while I am here:

- `web/tests/` is an existing location (`rss-default.test.mjs` is already on `dev`), so the `.test.mjs` files land where they belong. No concern there.
- These three change what a person sees, so per the convention in this repository they wait for a maintainer to look at the running interface before merging, separately from CI being green. I have flagged them for that.

Maya-Kid:
Completed the requested branch alignment: #854 → #856 → #855. Pagination is now replayed on #856 at `d01cc796b9ad2e3911fd56cb8133763c2b121b8c`, with the import/i18n conflicts resolved. All three contain current dev `474b904`; the PR description now states the stacked dependency explicitly.

The stack passes 116 frontend module tests, 27 browser subtests and build/typecheck/guard. The combined tree also passes the actual Linux/PostgreSQL 65-conversation pagination scenario and the other four backend/browser scenarios. This does not replace your requested visual review. Full verification and historical evidence are in the updated descriptions.

Maya-Kid:
Completed the post-merge synchronization of the remaining UI stack onto `dev@7cfeedf`, preserving the requested #854 → #856 → #855 order. Current heads: #854 `bd11c1d9ea61f5015d55ed8bc0f656c1bb2e2717`, #856 `ba964f3aaa94ae514f9e585c04329f642708299d`, #855 `a6737587017c740a172dd2fba21bf9d55ab5d497`. The i18n conflict resolution retains the merged stream-interruption strings together with the history/pagination strings.

The complete UI tree is identical to the previously validated integration tree. Revalidation passes 129 frontend module tests, 27 browser subtests (29 including parent tests), typecheck/guard/build. Updated all three descriptions; maintainer visual review remains separate and outstanding.

WaylandYang (dismissed):
Reviewed as the top of the stack, so only the +564 on top of #856: the sidebar moves
to `useInfiniteQuery` over the existing `limit`/`offset` list endpoint (the server
already returns `total`; `api.ts` already typed it), dedupes by id, and adds a "load
earlier" control with an error/retry state. The query key gains `convSearch` and is
gated on `kb?.id === kbId` like the rest of the page after #854, and `invalidateList`'s
prefix key still matches, so a rename or delete refetches the loaded page range
rather than appending a stale offset.

One property worth writing down where the dedupe comment is: offset pagination can
*skip* as well as duplicate. If a conversation moves to the top between two page
fetches (a new message in an old thread), everything below it shifts by one and the
item at the old boundary is never fetched. Dedupe handles the duplicate; only the
invalidation-driven refetch heals the skip, and only after something invalidates. For
a sidebar that is acceptable, and the fix (keyset on `updated_at, id`) is a server
change, not this PR's, but the comment currently reads as if dedupe covers both.

Three new i18n keys in both packs. The 530 test lines are on-demand browser probes,
not CI, as with the rest of the stack. LGTM; merge after #856.

WaylandYang:
Rebased onto `dev` as a maintainer edit, same as #856: #854 and #856 landed as squashes (9fa75e4, 315f46c), so the four earlier stack commits here conflicted with their own squashed copies. The branch is now exactly your pagination commit (08728f5 → e0e2f01) on current `dev`; authorship and sign-off unchanged; `pnpm guard` / `typecheck` / `test` (132) / `build` pass. If you have local work here, `git fetch && git reset --hard origin/fix/chat-conversation-pagination` first.

PR #856, opened 2026-09-21
Refresh persisted chat history once after an idle reattachment

Maya-Kid:

Based on `dev@474b904dc3e8014d42404a3c067af9906036a953`; current head `a155bbab6465e2e6a454bd53a98c931147c29951`.

Merge order follows the maintainer’s request: **#854 → #856 → #855**. This branch includes #854 as its prerequisite.

---

Depends on #854 (view-request ownership). This branch includes that prerequisite so stale refreshes cannot overwrite a later navigation or send. The additional commits for this behavior are `cb82751` and `9afbe39`; after #854 merges, GitHub's diff will reduce to those changes.

An answer can commit after the browser reads a user-only history but before it reattaches to the live producer. The server then correctly returns `idle`, while the page remains stuck with the unanswered question. On that initial idle handoff, read the persisted history once and apply it only while the initiating view request still owns the page. If it is still unanswered, display a bounded idle status; a failed read offers an explicit retry. This never re-POSTs the question or recursively reattaches.

Validation, based on `dev@ea0557b` plus #854, final head `9afbe39`:
- Four new history-handoff browser regressions fail before this change while the 11 prerequisite cases pass.
- All 16 browser subtests pass after the fix: the 11 prerequisite cases plus completed-between-reads, bounded user-only history, late refresh after navigation, retry after refresh failure, and a new send taking ownership before the old refresh arrives. The tests render the real Chat/router/query/live-answer implementation with controlled HTTP responses.
- Frontend 116 module tests and build/typecheck/guard pass; the final additional send-race case and typecheck pass as well.
- Chrome with the actual Linux backend and PostgreSQL held reattachment until a locally scripted model's answer had committed and the producer had unregistered. The real `idle` response caused exactly one extra detail GET, rendered the saved answer, and caused zero chat POSTs.
- Integration tree `f7bcd0878bf490d6efb122b71d6f13d379df4709`, including #845 and #851's strict stream terminal handling, passes Rust fmt, strict workspace Clippy, 1,013 tests (one pre-existing external-HTTPS RSS test ignored), workspace build, 129 frontend module tests, 27 browser subtests and frontend build.

Run `node --test tests/chat-view.test.mjs` from `web`, with optional `CHAT_PLAYWRIGHT_PATH` and `CHAT_CHROMIUM_PATH`. No new dependency, polling loop, model retry, or server generation policy is introduced. All model integration evidence uses a local scripted endpoint, not a paid model.


WaylandYang (dismissed):
Reviewed as the second commit pair of the stack (#854 ⊂ #856 ⊂ #855), so only the
+231 on top of #854.

The idle handoff is the right shape: `reattachChat` answering `idle` means nothing is
running *now*, which says nothing about whether the answer committed between the
history read and the attach, and the one re-read closes exactly that window. The
guards are the ones that matter: `checkedIdle` so it can never recurse, `handle`
null so a snapshot that did arrive wins, `ownsView(owner)` before and after the
await so a navigation or a new `send()` (which claims a new owner and clears
`idleHistoryKey`) makes the late read a no-op. `historyTurns` pulled out so the two
readers can't drift is a good small move.

Two things to know, neither blocking:

- This is one extra `GET .../detail` on every open of a conversation whose last turn
  is the user's, whether or not anything was ever running. That is the honest cost
  of not trusting the first read; fine at chat scale.
- The "no active answer" line keys on `idleHistoryKey === loadedKey === viewKey(...)`
  and `!streaming`, so it disappears the moment a new send begins. Good; just noting
  it is the only place the flag is read, in case someone later reuses it.

Same caveat as #854: the 197 test lines are on-demand browser probes, not CI. LGTM.

WaylandYang:
Rebased this branch onto `dev` as a maintainer edit so it can merge: #854 landed as a squash (9fa75e4), which left the two original #854 commits here conflicting with their own squashed copy. The branch is now exactly your two commits (083f4ed → 192a1fa, 1881004 → cd87e88) on top of current `dev`; authorship and sign-offs are unchanged, and `pnpm guard` / `typecheck` / `test` (132) / `build` pass on the result. If you have local work on this branch, `git fetch && git reset --hard origin/fix/chat-history-reattach-handoff` before continuing. #855 will get the same treatment after this merges.

